PR #5, opened 2020-05-27
Correlation of genes in different partitions

ajlee21:
This PR partitions the dataset into PAO1 and PA14 only samples and examines what the correlation structure looks like between genes.

The main changes are in NB 2_explore_corr_of_genes.ipynb


pilot_experiment/2_explore_corr_of_genes.ipynb
```diff
@@ -1397,188 +1405,293 @@
...
+    "                 rect=[0, 0.03, 1, 0.95])"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "**Some sanity checks**\n",
+    "**Note about visualizations:**\n",
+    "* Based on the density plot, we observed a shift in the accessory-accessory gene correlation scores. This density plot represents the probability of a random variable falling within a particular range of values (P(0 <= X <= 0.5)). This probability is given by the integral of this variable's PDF over that range -- that is it is given by the area under the density function. So there is a higher likelihood of correlation scores > 0.5 for accessory-accessory genes compared to core-core genes. But the *exact probability* of accessory-accessory genes having a correlation score > 0.5 is not known from just visually inspecting the density plot. To get the exact probability we would need to calculate the integral from 0.5 onward.\n",
```
jjc2718, line 1538: But you could approximate this integral in this case, right? Since your density plot is just a continuous approximation of your discrete data (gene correlations), you could just count the number of gene correlations > 0.5 and divide it by the total number of gene correlations.

pilot_experiment/2_explore_corr_of_genes.ipynb
```diff
@@ -1397,188 +1405,293 @@
...
+    "                 rect=[0, 0.03, 1, 0.95])"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "**Take away**:\n",
+    "* If high corr(acc, acc) was due to gene absence, then I would’ve expected the average expression of accessory genes to be tightly spread near 0. But instead we see that average gene expression of accessory genes is similar to the core gene expression, even using only PA14 strains\n",
+    "* What can explain this? "
```
jjc2718, line 1694: Yeah, this is interesting. To me, it doesn't look like the expression of the accessory genes is substantially different from the core genes, maybe slightly lower but definitely not close to 0 in most cases.

As an aside, what are the units used to quantify gene expression here? Since this is from ADAGE this is microarray data, right?

pilot_experiment/2_explore_corr_of_genes.ipynb
```diff
@@ -1397,188 +1405,293 @@
...
+    "             color='grey',\n",
+    "             hist = False, \n",
+    "             kde = True,\n",
+    "             kde_kws = {'shade': True}\n",
+    "            )\n",
     "\n",
     "plt.legend(prop={'size': 12})\n",
-    "plt.title('Density of correlation scores per group')\n",
-    "plt.ylabel('Density')"
+    "plt.title('Probability density of correlation scores per group')\n",
```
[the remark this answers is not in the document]
ajlee21: Good point, I will add this

pilot_experiment/2_explore_corr_of_genes.ipynb
```diff
@@ -1397,188 +1405,293 @@
...
+    "            )\n",
+    "sns.distplot(shuffled_all_corr_score,\n",
+    "             label='shuffled', \n",
+    "             color='grey',\n",
+    "             bins=bins_corr,\n",
+    "             kde=False,\n",
+    "             ax=axes[1,1]\n",
+    "            )\n",
+    "\n",
+    "plt.suptitle('Histogram of correlation scores per group')\n",
```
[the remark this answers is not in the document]
ajlee21: I see what you're saying. I think at this stage we're able to see the difference by plotting them individually and looking more at the density curves so I will leave it for now.

pilot_experiment/2_explore_corr_of_genes.ipynb
ajlee21: Yes that would be an approximation you could do for future.

pilot_experiment/2_explore_corr_of_genes.ipynb
ajlee21: Very good observation. So this data was using arrays to quantify expression which is why there are no 0 expressed values for those absent genes. This is motivating our next step, which is to use rna-seq instead

pilot_experiment/2_explore_corr_of_genes.ipynb
```diff
@@ -1397,188 +1405,293 @@
...
-    "**Some sanity checks**\n",
+    "**Note about visualizations:**\n",
+    "* Based on the density plot, we observed a shift in the accessory-accessory gene correlation scores. This density plot represents the probability of a random variable falling within a particular range of values (P(0 <= X <= 0.5)). This probability is given by the integral of this variable's PDF over that range -- that is it is given by the area under the density function. So there is a higher likelihood of correlation scores > 0.5 for accessory-accessory genes compared to core-core genes. But the *exact probability* of accessory-accessory genes having a correlation score > 0.5 is not known from just visually inspecting the density plot. To get the exact probability we would need to calculate the integral from 0.5 onward.\n",
+    "* While the shift if very clear to see in the density plot, the meaning of the y-axis is not as intuitive, so we also plot histograms for each group of genes. The histogram also shows a shift in the accessory-accessory gene correlation scores and also shows that the number of accessory-accessory interactions is orders of magnitude lower compared to core-core interactions.\n",
```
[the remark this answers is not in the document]
ajlee21: To me, if we calculate the skewness of each distribution, this will give a quantification for how different the correlation distributions are from the shuffled dataset. So it will quantify that we have some signal in our data. But we're mainly interested in the shift between the accessory-accessory (blue) compared to the other distributions.

If I run the skewtest i get the following results
```
core-core skewness:  SkewtestResult(statistic=22.532166392187893, pvalue=2.008873119085481e-112)
acc-core skewness:  SkewtestResult(statistic=-14.491763183083687, pvalue=1.3659105195229116e-47)
acc-acc skewness:  SkewtestResult(statistic=78.40894732416196, pvalue=0.0)
shuffled skewness:  SkewtestResult(statistic=175.67273941219565, pvalue=0.0)
```
I'm not sure why core-core and shuffled are coming back with significant skewness.

PR #10, opened 2020-08-05
Add DE validation analysis

ajlee21:
The main change in this PR is to add a notebook (`DEG_validation.ipynb`) whose goal is to assess that the salmon quantification is working as expected. This PR uses a small dataset of PAO1 and PA14 samples and performs a DE analysis using DESeq and compares the findings to the publication from which they originated.

Ideally I wanted to run this validation using the dataset I used for the correlation analysis, however the associated publication did not run a DE analysis.


sra_experiment/nbconverted/DEG_validation.py
```diff
@@ -0,0 +1,311 @@
...
+# In[19]:
+
+
+get_ipython().run_cell_magic('R', '', '# Select 59\n# Run one time\n#if (!requireNamespace("BiocManager", quietly = TRUE))\n#    install.packages("BiocManager")\n#BiocManager::install("DESeq2")')
+
+
+# In[20]:
+
+
+get_ipython().run_cell_magic('R', '', '# Load the DESeq2 library\nlibrary("DESeq2")')
```
jjc2718, line 225: If you want, you can wrap the `library` call in a `suppressPackageStartupMessages` call to get rid of these warnings that happen when you load an R package. I don't think these warnings are particularly helpful, so probably fine to just suppress them.

sra_experiment/nbconverted/DEG_validation.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+# Check ordering of sample ids
+utils.check_sample_ordering(expression_data_file, metadata_file)
+
+
+# In[23]:
+
+
+get_ipython().run_cell_magic('R', '-i metadata_file -i expression_data_file -i out_file', "\nsource('../core_acc_modules/DE_analysis.R')\n\nget_DE_stats_DESeq(metadata_file,\n                   expression_data_file,\n                   out_file)")
```
jjc2718, line 247: Here you can use `suppressMessages` to get rid of the progress messages if you want (although these ones are slightly more useful)

sra_experiment/nbconverted/DEG_validation.py
```diff
@@ -0,0 +1,311 @@
...
+# with those genes found using DESeq
+top_degs_by_padj = selected_DE_stats.index
+len(set(top_degs_by_padj).intersection(genes_found_from_GE))/len(genes_found_from_GE)
+
+
+# **Conclusions:**
+# 
+# Our DE analysis found ~1.1K significantly differentially expressed genes
+# 
+# (Check 1): The DEGs identified using DESeq (\~1.1K genes) is fairly consistent with the genes that were 1 standard deviation outside the correlation threshold (\~200 genes) -- there was a 75% overlap in these gene sets. This very roughly validates that DESeq is working as expected. I wouldn't expect the numbers to be this different though.
```
jjc2718, line 302: This makes sense to me. DESeq is using a probability model that should (at least in theory) be a good match for the way RNA-seq data is generated, so I would expect it to have better statistical power to detect DE genes than your quick sanity check. The fact that what you found is (roughly) a subset of what DESeq found makes me fairly confident that this is working correctly.

sra_experiment/nbconverted/DEG_validation.py
```diff
@@ -0,0 +1,311 @@
...
+# 
+# (Check 2) The number of DEGs identified (\~1.1K genes) using DESeq is fairly consistent with the number of differentially expressed genes found in [Sana et. al](https://jb.asm.org/content/201/21/e00362-19) (\~2K genes). We also spot checked specific genes that were found. We found the 4 genes highlighted in the Sana et. al. publication, including the main qsIA gene (PA1244) that the paper found to be more highly expressed in PAO1 vs PA14. Difference are likely due to differences in the package used.
```
jjc2718, line 304: Nice!

sra_experiment/nbconverted/DEG_validation.py
```diff
@@ -0,0 +1,311 @@
...
+import seaborn as sns
+import matplotlib.pyplot as plt
+import numpy as np
+from core_acc_modules import utils, paths
+from rpy2.robjects import pandas2ri
+pandas2ri.activate()
+
+
+# ### Download data for validation
+# Data from [Sana et. al](https://jb.asm.org/content/201/21/e00362-19) found ~ 2K DEGs between 2 strains where QS genes were DEGs.
```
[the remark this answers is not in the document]
ajlee21: Ah, yes good point.

sra_experiment/nbconverted/DEG_validation.py
```diff
@@ -0,0 +1,311 @@
...
+# In[3]:
+
+
+get_ipython().run_cell_magic('bash', '', 'mkdir $paths.FASTQ_TEST_DIR\nfastq-dump $paths.SRA_DIR/SRR8486287.sra --outdir $paths.FASTQ_TEST_DIR/\nfastq-dump $paths.SRA_DIR/SRR8486288.sra --outdir $paths.FASTQ_TEST_DIR/\nfastq-dump $paths.SRA_DIR/SRR8486289.sra --outdir $paths.FASTQ_TEST_DIR/\nfastq-dump $paths.SRA_DIR/SRR8486290.sra --outdir $paths.FASTQ_TEST_DIR/')
+
+
+# In[4]:
+
+
+get_ipython().run_cell_magic('bash', '-s $paths.PAO1_QUANT_TEST $paths.FASTQ_TEST_DIR $paths.PAO1_INDEX', 'mkdir $1\n\nfor FILE_PATH in $2/*;\ndo\n\n# get file name\nsample_name=`basename ${FILE_PATH}`\n\n# remove extension from file name\nsample_name="${sample_name%_*}"\n\n# get base path\nbase_name=${FILE_PATH%/*}\n\necho "Processing sample ${sample_name}"\n\nsalmon quant -i $3 -l A \\\n            -r ${base_name}/${sample_name} \\\n            -p 8 --validateMappings -o $1/${sample_name}_quant\ndone')
```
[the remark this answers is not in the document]
ajlee21: Still getting used to bash:) Thanks

core_acc_modules/utils.py
```diff
@@ -214,3 +214,28 @@ def permute_expression_data(input_data):
...
+    # Check ordering of sample ids is consistent between gene expression data and metadata
+    metadata = pd.read_csv(metadata_file, sep="\t", header=0, index_col=0)
+    metadata_sample_ids = list(metadata.index)
+
+    expression_data = pd.read_csv(expression_file, sep="\t", header=0, index_col=0)
+    expression_sample_ids = list(expression_data.index)
+
+    if metadata_sample_ids == expression_sample_ids:
+        print("sample ids are ordered correctly")
+        return
```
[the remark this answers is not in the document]
ajlee21: Thanks. Still learning how to use functions and I reflexively thought I needed to have a return

sra_experiment/nbconverted/DEG_validation.py
```diff
@@ -0,0 +1,311 @@
...
+
+# Get raw read counts using PAO1 reference
+# Read through all sample subdirectories in quant/
+# Within each sample subdirectory, get quant.sf file
+data_dir = paths.PAO1_QUANT_TEST
+
+expression_data = pd.DataFrame(
+    pd.read_csv(file, sep="\t", index_col=0)["NumReads"].
+    rename(file.parent.name.split("_")[0]) 
+    for file in data_dir.rglob("*/quant.sf"))
```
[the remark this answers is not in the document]
ajlee21: good point

sra_experiment/nbconverted/DEG_validation.py
```diff
@@ -0,0 +1,311 @@
...
+expression_data = expression_data.reindex(columns=core_pao1_genes)
+print(expression_data.shape)
+expression_data.head()
+
+
+# In[8]:
+
+
+# Convert values to integers for DE analysis
+# Not sure why the "Numreads" values are floats
```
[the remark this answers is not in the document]
ajlee21: You're correct. Salmon produced estimated counts but DESeq requires integers so there is some inconsistency there, but looks like its a known problem. Looks like if I had used tximport (which maps transcript-level expression to gene-level expression), it would have returned counts, but since I was already at the gene-level I skipped this step and thus the estimated counts that I manually rounded. But this brings up a good point that I need to look more carefully to see what tximport and DESeq are doing in terms of processing the data to make sure my rounding isn't doing something it shouldn't

sra_experiment/nbconverted/DEG_validation.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+sns.scatterplot(data=pao1_v_pa14_df, x='pao1_mean', y='pa14_mean')
+plt.plot([0,140000],[0,140000])
+
+
+# In[18]:
+
+
+# Rough calculation of the number of genes that differ
```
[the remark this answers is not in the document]
ajlee21: I had trouble trying to describe this well. I will add more to this comment to explain

sra_experiment/nbconverted/DEG_validation.py
ajlee21: I had no idea about this! So nice

sra_experiment/nbconverted/DEG_validation.py
ajlee21: Yea i agree...this i think I'll leave for now since there is not an overwhelming number of messages and they are giving some information about the run

sra_experiment/nbconverted/DEG_validation.py
ajlee21: I agree...I was a little panicked because I thought they should still be closer in range, but they do appear to be a subset. I'll see if things change once i look more into tximport and DESeq and how they are processing the data (i.e. scaling between samples)

PR #11, opened 2020-12-15
Test phage reference genome

ajlee21:
This PR test phage reference genome locally before this reference is added to the processing pipeline to run on DISCOVERY (Dartmouth computing core).

Files to review are in `test_phage_reference/`:
1. `0_process_phage_genome.ipynb`: short notebook to filter downloaded phage genome data to remove non-pseudomonas entries
2. `1_align_quantify_samples.ipynb`: preliminary data to test alignment against phage reference genome. Use samples we expect to align well and those that we do not expect to align well.
3. `2_validate_expression_data.ipynb`: look at the resulting expressions to see how they look

Based on this small test. It does not look like phage samples align to the phage reference, as we would have expected. This was a first pass experiment. This is likely because I am accidentally using phage genome sequences as input instead of phage transcriptomes. This will be the next thing to try. Another possible thing to try, is to use phage genes instead of the entire genome.


test_phage_reference/nbconverted/1_align_quantify_samples.py
```diff
@@ -0,0 +1,211 @@
+
+# coding: utf-8
+
+# # Aign and quantify samples
```
arielah, line 4: You probably mean "align". Feel free to slap me next time you see me for pointing out typos this minor.

test_phage_reference/nbconverted/1_align_quantify_samples.py
```diff
@@ -0,0 +1,211 @@
...
+
+np.random.seed(123)
+
+
+# ### Setup SRA toolkit -- only needs to be run once
+
+# In[2]:
+
+
+# Download latest version of compiled binaries of NCBI SRA toolkit
```
arielah, line 28: Is this only going to be run in notebook form? Because if a future user tries to run these scripts as a plug-and-chug, they may encounter problems with the download info being commented out. But if that scenario's never going to happen don't worry about it.

test_phage_reference/nbconverted/1_align_quantify_samples.py
```diff
@@ -0,0 +1,211 @@
...
+# ### Obtain a transcriptome and build an index
+# 
+# Here we are using [Salmon](https://combine-lab.github.io/salmon/)
+# 
+# **Input:**
+# * Target transcriptome
+# * This transcriptome is given to Salmon in the form of a (possibly compressed) multi-FASTA file, with each entry providing the sequence of a transcript
+# * We downloaded the phage GENOMES from NCBI GenBank
+# 
+# **Note:** For prokaryotes, transcripts and genes have a more 1-1 mapping so we're using genes for our reference transcriptome and so we don't need to use tximport to map transcript quants to genes.
```
arielah, line 101: This is very helpful documentation!

test_phage_reference/nbconverted/1_align_quantify_samples.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+# Get phage index
+get_ipython().system(' salmon index -t $paths.PHAGE_REF -i $paths.PHAGE_INDEX')
+
+
+# ### Quantify gene expression
+# Now that we have our index built and all of our data downloaded, we’re ready to quantify our samples
+# 
+# **Input:**
```
arielah, line 131: If you have to put in multiple "input" "output" comments into a script, that violates the rule I was taught of having each script do one thing and do it well. I can understand combining them in a notebook environment, but I think building the index is different enough it should be a separate script imho.

test_phage_reference/nbconverted/1_align_quantify_samples.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+if not os.path.exists(paths.PHAGE_QUANT):
+    os.makedirs(paths.PHAGE_QUANT)
+
+
+# In[12]:
+
+
+get_ipython().run_cell_magic('bash', '-s $paths.PHAGE_QUANT $paths.FASTQ_DIR $paths.PHAGE_INDEX', '\nfor FILE_PATH in $2/*;\ndo\n\n# get file name\nsample_name=`basename ${FILE_PATH}`\n\n# remove extension from file name\nsample_name="${sample_name%_*}"\n\n# get base path\nbase_name=${FILE_PATH%/*}\n\necho "Processing sample ${sample_name}"\n\nsalmon quant -i $3 -l A \\\n            -1 ${base_name}/${sample_name}_1.fastq \\\n            -2 ${base_name}/${sample_name}_2.fastq \\\n            -p 8 --validateMappings -o $1/${sample_name}_quant\ndone')
```
arielah, line 170: Now this is just an opinion thing, but is there a reason not to just make these steps a bash script? It seems unwieldy to have all of this in a python command, especially with the '\n' stuff.

test_phage_reference/nbconverted/2_validate_expression_data.py
```diff
@@ -0,0 +1,100 @@
...
+expression_pao1_df.head()
+
+
+# ## Negative case
+
+# In[16]:
+
+
+# Check there is no expression in E. Coli sample using phage reference
+sns.distplot(expression_phage_df.loc["SRR12922100"], kde=False)
```
arielah, line 40: This is fancy

arielah (commented):
Very nicely documented pull request, as always! My comments are all matters of preference, as someone who almost never uses notebooks, so feel free to ignore if you want your setup to stay notebook-centric.

test_phage_reference/nbconverted/1_align_quantify_samples.py
ajlee21: Yes I agree. The index only needs to be built once but the quantification can run multiple times per sample and index so I'll separate this out.

test_phage_reference/nbconverted/1_align_quantify_samples.py
ajlee21: Your intuition is right on! This should be a bash script. These are meant to be some quick test scripts to see what's going on with the data and use this information to update the working pipeline that is on Discovery. Since I'm more familiar with python this is what I defaulted to when I was quickly trying to put together some tests. But these will eventually be run as scripts on a cluster

test_phage_reference/nbconverted/1_align_quantify_samples.py
ajlee21: Good point. I have separated out the scripts to setup the analysis from those that run the analysis for future use. But this really is just a quick test so I don't anticipate a user using this. However the notebooks were more of a way for me to organize my thoughts on how this pipeline is working.

ajlee21:
Thank you for all your helpful comments! Please let me know if you have additional comments/suggestions @arielah

arielah (approved):

PR #12, opened 2020-12-22
Continue testing phage reference genome

ajlee21:
This PR continues testing the phage reference genome. Previously we found that we didn't get a good alignment of phage samples against our phage reference genome, which was unexpected. It turned out that I was accidentally using genome samples instead of transcriptome samples. Here we are using transcriptome samples and found:

1. Aligning E. Coli sample against Pseudomonas PAO1, PA14 and phage reference genomes, most genes were found to have 0 expression, as expected
2. Aligning PAO1 sample against the Pseudomonas PAO1 and phage reference genomes, most genes in PAO1 sample are nonzero, including phage gene that the sample is known to contain -- as expected
3.  Aligning PA14 sample against the Pseudomonas PA14 reference genome, most genes in PA14 sample are nonzero, including phage gene that the sample is known to contain -- as expected

Overall these test results are very promising, but there were some additional observations that I'm not sure how to explain and will be discussed during my weekly meeting. These observations include the PAO1 and PA14 genes having TPM values < 1000 and outlier genes with > 300K TPM.

PR #13, opened 2021-01-14
Add details about how Salmon works

ajlee21:
This PR is following up on some of the comments made in https://github.com/greenelab/core-accessory-interactome/pull/12

I have updated the comments in the notebooks to add more clarity. The changes (fairly minor) include:
1. Added a readme to describe the analysis that is being performed by these notebooks in `test_phage_reference/`
2. Updated comments in notebooks to describe how Salmon is working and try to address some of the outputs observed.
3. Updated Salmon version and parameter settings to be consistent with what is currently implemented in the pipeline we'll use to process samples.
4. Re-organized files to split download from quantification step
5. Updated fastq and quant calls to be consistent with what is currently implemented in the pipeline:  https://github.com/hoganlab-dartmouth/pa-seq-compendia/blob/main/salPa_byx.sh

After learning more about Salmon, I think I have a bit of a better sense for what is happening, but I still have some questions about how you're processing the PAO1 and PA14 compendia to make sure I'm being consistent with your implementation while also applying Salmon appropriately @georgiadoing:
* When building the index I'm getting that PAO1 34 duplicates are removed, PA14 37 duplicates are removed, Phage  391 duplicates removed. Are these duplicates expected? Are you keeping duplicates?
* During our last meeting you had a comment about adjusting the k length for building the index. My understanding is that the k will determine the minimum length of valid matches for the quasi-mapping step. So a smaller k can improve sensitivity but can possibly also lead to ambiguous mapping? I think you're using the default k=31, is that correct? 
* Since I'm using phage genomes for the reference, I'm getting warnings about the length of the entries. Do you know if Salmon is still including these in the index?
* Its looks like you're quantifying all fastq files associated with a specific SRR#. In my case I have paired-end samples, so it looks like Salmon is treating them independently. I'm not sure if this is what we want. I remember Jake mentioning that he reversed the R strand and treated the F,R independently but I don't see that pre-processing step in yours so I just wanted to check.


test_phage_reference/2_download_data.ipynb
```diff
@@ -0,0 +1,179 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Download data\n",
+    "\n",
+    "This notebook is downloading data from SRA and then using the module, `fasterq-dump`, from the SRA toolkit to get the fastq files associated with the downloaded SRA files.\n",
+    "\n",
+    "Note: Need to delete `sra` folder between runs otherwise `fastq-dump` will be called on all files in `sra` folder which can include more than your sra accessions."
```
rando2, line 11:
```suggestion
    "Note: Need to delete `sra` folder between runs; otherwise, `fastq-dump` will be called on all files in `sra` folder which can include more than your sra accessions."
```

rando2 (approved):
This seems reasonable to me @ajlee21! At some point I'm happy to test it and make sure all of the explanations help me make sense of what's being generated -- just let me know if that would be helpful!

ajlee21:
Thank you for reviewing this PR @rando2. The next PR will be to re-process the data, and it would definitely be helpful to get feedback on the output generated and if my comments make sense for that one.

PR #15, opened 2021-02-02
Add exploratory analysis

ajlee21:
This PR adds several analyses to "look at the data" before diving into the eventual analysis of computing the correlation of expression profiles between the different gene groups (core, accessory):

Changes include:
1. Add `clustering_by_metadata.ipynb` to examine how samples are clustered by metadata. 
2. Add `DE_analysis_PRJNA283002.ipynb` performed DE analysis of one of the experiments to confirm that we are getting consistent results compared to the publication. Results look consistent.
3. Add `expression_activity.ipynb` to visualize the distribution of the expression data and confirm that accessory genes are lowly expressed if at all.

Using raw TPM values the clustering looked like there is predominately clustering based on the biology that we care about (PAO1 vs PA14) but there is still some variance contributed by platform. For now I think TPM normalization appears sufficient for our purposes as pilot, but this may need to be revisited in the future:
image

image


ajlee21:
Having some conda environment issues that I cannot figure out. I created a new environment after adding one new package but when I tried to re-run the clustering notebook (which had previously run successfully), I got the error message:
`
Cannot interpret '<attribute 'dtype' of 'numpy.generic' objects>' as a data type plotnine
`
I assume there is some dependency that got updated with the new environment but couldn't track it down. So for now I am using an older conda env. However even if I match the revert the yml file to this old version, I'm still getting this error.

core_acc_modules/DE_helper.py
```diff
@@ -0,0 +1,107 @@
+"""
+Author: Alexandra Lee
+Date Created: 18 December 2020
```
ben-heil, line 3: Is there a reason that this is backdated two months? Is it from a different repo, or has it just been in the works for awhile?

core_acc_modules/DE_helper.py
```diff
@@ -0,0 +1,107 @@
...
+        # Get samples ids to remove
+        samples_to_remove = list(
+            process_metadata[process_metadata["processing"] == "drop"].index
+        )
+
+        # Remove samples
+        expression = expression.drop(samples_to_remove)
+
+    # Cast as int
+    expression = expression.astype(int)
```
ben-heil, line 76: Why is expression getting cast to an integer here? Are you working with raw counts?

explore_data/nbconverted/DE_analysis_PRJNA283002.py
```diff
@@ -0,0 +1,145 @@
...
+]
+
+
+# In[12]:
+
+
+selected_DE_stats.loc[published_DEGs]
+
+
+# The differential expression results can be found in [Figure 1](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4524035/) of the paper. Spot checking it looks like the genes have consistent direction of logFC.
```
ben-heil, line 145: Are the NAN values in this cell expected?

explore_data/nbconverted/cluster_by_metadata.py
```diff
@@ -0,0 +1,277 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# # Clustering by metadata
+# 
+# This notebook examines how the data clusters by metadata. We would expect that the data cluster by strain type (PAO1 and PA14).
+# 
+# Other metadata?
```
ben-heil, line 8: It's unclear what this line means

explore_data/nbconverted/cluster_by_metadata.py
```diff
@@ -0,0 +1,277 @@
...
+    legend_key=pn.element_rect(fill='white', colour='white'), 
+    legend_title=pn.element_text(family='sans-serif', size=15),
+    legend_text=pn.element_text(family='sans-serif', size=12),
+    plot_title=pn.element_text(family='sans-serif', size=15),
+    axis_text=pn.element_text(family='sans-serif', size=12),
+    axis_title=pn.element_text(family='sans-serif', size=15)
+    )
+fig += pn.guides(colour=pn.guide_legend(override_aes={'alpha': 1}))
+
+print(fig)
```
ben-heil, line 135: Is there a reason that this data is so colinear in PCA space? That is to say, do you have an intuition for why the PA01 points come off in spikes?

ben-heil: Like the PA14 points look like I would expect something in PCA space to look, but the PA01 points feel like something is breaking

ben-heil (approved):
Looks good! Sorry to hear that the new version of one of the libraries isn't playing nicely with others, but I guess that's a win for using Anaconda

core_acc_modules/DE_helper.py
ajlee21: This was originally written in another repo. I'm not sure what the convention is for dates. Do you think it makes more sense to note where it was originally from and keep the date as is?

core_acc_modules/DE_helper.py
ajlee21: Great catch! DESeq is expecting counts and i accidentally have TPM values so I will need to re-process the data.

explore_data/nbconverted/cluster_by_metadata.py
ajlee21: This was me thinking "outloud" of other metadata fields to consider

explore_data/nbconverted/cluster_by_metadata.py
ajlee21: Very good observation. I also was not expecting this shape. I'm not sure what could be going on here so something I'll need to spend some more time exploring.

explore_data/nbconverted/DE_analysis_PRJNA283002.py
ajlee21: NaN's occur if the samples are all 0 which is the case for 2 of the genes. But I'm not sure what explains the third case and need to look into this. There are fitlering parameters that can result in Nans (as explained in the link), but I have turned these off.

https://bioconductor.org/packages/release/bioc/vignettes/DESeq2/inst/doc/DESeq2.html#why-are-some-p-values-set-to-na

PR #18, opened 2021-03-19
Explore data

ajlee21:
This PR is continuing to explore (following up on the previous [PR](https://github.com/greenelab/core-accessory-interactome/pull/16)) the input data that we are using for network analysis.

In particular we examined:
1. Clustering of samples based on core and accessory gene expression
2. Examined samples that had high expression of both PAO1 and PA14 accessory genes
3. Proportion of samples with low expression


explore_data/cluster_by_accessory_gene.py
```diff
@@ -0,0 +1,609 @@
...
+# Drop label columns
+pao1_expression_numeric = pao1_expression_label.drop(["Strain type"], axis=1)
+pa14_expression_numeric = pa14_expression_label.drop(["Strain type"], axis=1)
+
+# +
+# Embed expression data into low dimensional space
+pca = PCA(n_components=2)
+# model_pao1 = umap.UMAP(random_state=123).fit(pao1_expression_numeric)
+model_pao1 = pca.fit(pao1_expression_numeric)
+pao1_encoded = model_pao1.transform(pao1_expression_numeric)
```
nrosed, line 258: in the PCA fit are the values normalized in anyway? Looking at the PCA plot, I'm wondering if it's being driven by a few genes that have very high expression.

explore_data/cluster_by_accessory_gene.py
```diff
@@ -0,0 +1,609 @@
...
+fig2 += pn.guides(colour=pn.guide_legend(override_aes={"alpha": 1}))
+
+
+print(fig2)
+# -
+
+# **Takeaway:**
+# * This is a positive control that core genes are expressed in both PAO1 and PA14 samples
+#
+# * Based on expression profiles of core genes in PAO1 vs PA14 samples, there doesn't look to be a clear clustering amongst PAO1 and PA14 samples. Is this expected?
```
nrosed, line 380: possibly, if the PCA is being driven by outlier expression after rescaling the data it could be that you see clusters. I would at least expect to see some structure... but I'm not very familiar with non-mammalian gene expression.

explore_data/cluster_by_accessory_gene.py
```diff
@@ -0,0 +1,609 @@
...
+
+# **Note:**
+# We can compare TPM between PAO1 and PA14 if they have similar total number of reads mapped. So perhaps we can compare PAO1 and PA14 compendia. See Georgia’s [google doc](https://docs.google.com/document/d/1OnvJNVkhK5ATnhHaeSmz8Ze_Iw9ixXhfSxAL_BSqbuM/edit)
+
+# **Takeaway:**
+# * This is a positive control that PAO1 annotated samples have higher median expression of PAO1-only genes compared to PA14-only genes. And similarly, PA14 annotated samples have higher median expression of PA14-only genes compared to PAO1-only genes. In other words, we expect that PA14-only genes will have either 0 or very low values in PAO1 samples and vice versa.
+#
+# * This result also shows that we can anticipate a very clear binning of our samples into PAO1 and PA14 if we use mapping rates.
+#
+# * Note: The NA strains are those where the strain information was not available in the metadata. By a quick manual spot check it looks like a bunch were clinical isolates (which is good since these NA seem to cluster with other clinical isolates).
```
nrosed, line 450: Why is the spread of expression values between PAO1 and PA14 different?

explore_data/cluster_by_accessory_gene.py
```diff
@@ -0,0 +1,609 @@
...
+]
+print(
+    f"Proportion of PAO1 samples with 0 median PAO1-only expression: {pao1_acc_expression_low.shape[0]/num_pao1_samples}"
+)
+pao1_acc_expression_low = pao1_acc_expression_label[
+    pao1_acc_expression_label["median acc expression_pao1"] < threshold
+]
+print(
+    f"Proportion of PAO1 samples with median PAO1-only expression < {threshold}: {pao1_acc_expression_low.shape[0]/num_pao1_samples}"
+)
```
nrosed, line 482: Why are you interested in the number of PA01 samples where the expression is below a threshold?

explore_data/cluster_by_accessory_gene.py
```diff
@@ -0,0 +1,609 @@
...
+    f"Proportion of PA14 samples with 0 median PAO1-only expression: {pa14_acc_expression_low.shape[0]/num_pa14_samples}"
+)
+pao1_acc_expression_low = pao1_acc_expression_label[
+    pao1_acc_expression_label["median acc expression_pa14"] == 0
+]
+print(
+    f"Proportion of PAO1 samples with 0 median PA14-only expression: {pao1_acc_expression_low.shape[0]/num_pao1_samples}"
+)
+# -
+
```
nrosed, line 528: Is there a way to make the above results into a confusion matrix?

explore_data/cluster_by_accessory_gene.py
```diff
@@ -0,0 +1,609 @@
...
+#
+# * PAO1 samples with high expression (>10 TPM) of PA14 specific genes
+#     * Sample names look like they might be switched: https://www.ncbi.nlm.nih.gov/sra?linkname=bioproject_sra_all&from_uid=517074
+# * PA14 samples with high expression (>10 TPM) of PAO1 specific genes
+#     * PA14 grown in blood: https://www.ncbi.nlm.nih.gov/sra/?term=SRX4326016
+#     * Some are mislabeled after checking publication: https://www.ncbi.nlm.nih.gov/sra/?term=SRX5099522, https://www.ncbi.nlm.nih.gov/sra/?term=SRX5099523, https://www.ncbi.nlm.nih.gov/sra/?term=SRX5099524
+#     * PA14 following treatment with antimicrobial manuka honey: https://www.ncbi.nlm.nih.gov/sra/?term=SRX7423386, https://www.ncbi.nlm.nih.gov/sra/?term=SRX7423388
+#
+# Based on this quick spot check, it does look like most cases where PAO1 samples with high PA14 gene expression or PA14 samples with high PAO1 expression appear to be due to mis-labelings. This supports our plans to auto-label samples based on mapping rates instead.
```
nrosed, line 609: Nice finding!

nrosed (approved):
Looks really nice! I only had a few questions in the file `explore_data/cluster_by_accessory_gene.py`, but other than that, looks promising!

explore_data/cluster_by_accessory_gene.py
ajlee21: This was more for me to get a better sense for where the expression is distributed, so it looks like most samples have a median expression < 50 TPM. This would allow be to quantify the distribution a bit

explore_data/cluster_by_accessory_gene.py
ajlee21: This is a great suggestion. I think this type of visualization will help make the interpretation easier.

explore_data/cluster_by_accessory_gene.py
```diff
@@ -0,0 +1,609 @@
...
+# -
+
+sample_to_strain_table.head()
+
+# What are the NA strains?
+print(sample_to_strain_table.shape)
+na_sample_ids = list(
+    sample_to_strain_table[sample_to_strain_table["Strain type"] == "NA"].index
+)
+print(len(na_sample_ids))
```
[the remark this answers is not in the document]
ajlee21: Ah yes. I keep forgetting about some of these really nice pandas functions!

explore_data/cluster_by_accessory_gene.py
ajlee21: These are great points. I think I'll try a couple of things here: 1) normalize and re-cluster and 2) remove other strain types and re-cluster

explore_data/cluster_by_accessory_gene.py
ajlee21: Good question. PAO1 higher median accessory expression than PA14. My initial guess that that maybe the PAO1 reference had higher mapping rates however, it looks like they have similar distributions of total reads mapped and mapping rate. So I'll have to think about what else might be causing this difference in range.

ajlee21: The good thing is that for the planned downstream analysis we're planning to separate between PAO1 and PA14 so this range shouldn't cause an issue in that sense

explore_data/cluster_by_accessory_gene.py
ajlee21: Looks like if I normalize and use UMAP to plot core gene expression of samples, there is some clustering of PAO1 vs PA14, which I would expect. This clustering isn't seen using PCs though, so I guess that means that more dimensions are needed for this separation.

ajlee21:
This was just a start at looking at the data. As I move forward I will continue to do more

PR #19, opened 2021-03-30
Create expression compendia based on mapping rates

ajlee21:
This PR adds a new directory to process data for use in downstream analysis. Specifically:

1. `0_decide_thresholds.ipynb` plots distribution of mapping rates to determine thresholds to use for mapping
2. `1_create_compendia.ipynb` creates compendia based on mapping rates
3. `2_validate_compendia.ipynb` visualizes binned data to confirm out method works as expected.

The plots are showing the median expression of PAO1 genes (PAO1 accessory genes) on the x-axis and the median expression of PA14-only genes (PA14 accessory genes) on the y-axis. Each point is a sample.

Here is using all the samples:
image

If we binned our samples accurately then for samples within our binned PAO1 compendium, we expect that samples will align along the PAO1-only axis. Similarly, for samples within our binned PA14 compendium, we expect that samples will align along the PA14-axis.

Here is the result using a mapping threshold of 30% and difference in mapping rate of 2%
image

image

If we increase the mapping rate and difference in mapping rate threshold, we start to get a more clear separation between PAO1 and PA14 samples, but the size of the compendia are limited

One alternative solution would be to use the median accessory expression to bin samples instead.

Note: Why are the number of PA14 samples much lower compared to PAO1? Would have thought that the mapping rate using the PA14 reference would be skewed toward 0, but the distribution of mapping rates are similar between PAO1 and PA14.


processing/0_decide_thresholds.py
```diff
@@ -0,0 +1,117 @@
...
+# * There is fairly rough bimodal distribution, so most samples align well to PAO1 reference or the PA14 reference.
+#
+# **Takeaway:**
+# * Based on the mapping rate distribution, looks like a mapping rate of 30% would be a good cutoff to identify those samples that have a high mapping to PAO1 or PA14
+# * Since PAO1 and PA14 have a larged shared core genome, we need to add an additional criteria to filter for those that are specific to PAO1 versus PA14. This additional criteria will use the difference in mapping rates
+#
+# A sample will be PAO1 if:
+# 1. PAO1 mapping rate >= 30%
+# 2. PAO1-PA14 mapping rate > 0%
+#
```
nrosed, line 116: What does it mean when the mapping rate is exactly 0?

processing/1_create_compendia.py
```diff
@@ -0,0 +1,255 @@
...
+    set(high_pao1_mapping_ids).intersection(high_pa14_mapping_ids)
+)
+
+print(len(high_pao1_pa14_mapping_ids))
+# -
+
+# **Some observations:**
+# * Looks like there are not any samples that map to both PAO1 and PA14 using our criteria
+# * The number of PA14 samples is much lower compared to PAO1. Does this mean that the mapping rates of PA14 samples mapped to PA14 reference lower?
+
```
nrosed, line 198: That's an interesting point. Looking at the mapping rate plot in `0_decide_thresholds`, it looks like the variance of the second hump in PAO1 and PA14 are slightly different, so the cutoff of 0.3 may be too strict for PA14? This may be overkill, but since you have the labels for some of the strains could you plot the mapping rates of those samples only so you can get a sense of the  "true" mapping rate distributions and then see what proportion of samples are outside of 30% mapping rate so you can maybe get an estimate of the FDR for that cutoff?

processing/1_create_compendia.py
```diff
@@ -0,0 +1,255 @@
...
+pa14_expression_label.head()
+
+# ## Quick comparison
+#
+# Quick check comparing our binned labels compared with SRA annotations
+
+pao1_expression_label["Strain type"].value_counts()
+
+pa14_expression_label["Strain type"].value_counts()
+
```
nrosed, line 244: What are the mapping rates of the misclassified strains?

nrosed: I'm just curious how close they are to the cutoff

processing/2_validate_compendia.ipynb
```diff
@@ -0,0 +1,2851 @@
...
...Aqf6B+Hr/QTjq8Atk10KjKD6XEANAuNGA66MiYfcPQHpIKNIDQ6BotYjNvoSIBrHQmM3eDpGIqliF7hS//PLLUBQFH3/8MW699VZ1+D/+8Q/MnDkTa9euRZ8+fbB+/Xr06dOn2oIlIqosRa+H3SEoNBlhEDv0DjvgcMBcXIgMxYhCh8Cs9b3kjzzTKQq6NW2KkIJC/PLbb3DkWxCTm4OmsdEI69WXbdGJ6qAK/ar37NmDDh06uCTEABAcHIzVq1cjKSkJ8+bNw8CBA7Fq1SoMGTKkWoIlIqosjdEPYY0bA2cvwqoAJpsNVo0WUDQwGYzw1XTYYrMj02qDRgFCdDoE6LRXnshHGP390b5TR7Rp3gyOwkJozP7QBQXV+ZcNiXxVhZpPZGdn47rrritz/IMPPoi33noLdrsdo0aNwjvvvFNlARIRVZWEbj2Q5KdDod4P6f6ByPfzQ6BOhxuuvx5+Gt9LdA5b8vHKbxew/FwaVqVm4MtLOciw2rwd1p+KotNBHxoGY3QM9MHBTIiJ6rAKJcXh4eFITU0tt8z48eOxbt06aLVaTJkyBUuWLKmSAImIqoqiKOg1eAgGN07ADWEhaBpVH13aJaFVw0SfS3ZSi4rxQVomMqxWGDQKcux2/GjJw4HcPLavJqojEhOr5ti2YsUKKIqC+fPnX3tQ5Th58iQURUHPnj2rdTllqVDziTZt2uDrr79GXl4e/P39yyw3cOBAbNmyBYMGDcKsWbMQERFRZYESEVUFjUaDVq1aoxVKetLxtWTY6VhBIfJsdtTT66HXKDBrNUgrtiK12AqbCAw+Wi9E5LsqdKc4OTkZhYWFePvtt69YtkePHvjyyy8RHh6OixcvXnOARETVxVcTYgBwOAR6RQNrqbvCdil5wUznw/VCRL6rQknx4MGDcfPNN2PHjh0Vmmm7du2wY8cOtGnTBvHx7BSfiOjPJtpoRLBOh2KHAzk2Oy4WW2HUKOgQaPbJLtiIiCqUFMfHx2Pnzp147733Kjzjpk2bYv/+/Thx4kSlgyMiourRyGxEr7BARBj0MCgKog16jIoMR0OzyduhEdUau3fvxuDBgxEREQGj0YjExET89a9/xblz59zKzp8/H4qiYMWKFR7n5an971dffQVFUTB58mSkpqZi+vTpaNCgAXQ6HRYtWlTpuDdu3IipU6fihhtuQFBQEPz9/dGmTRs
```
nrosed, line 2851: Plots look nice! Is there some commonality between the clustered samples? For example, are they from the same experiment?

nrosed (approved):
Looks good! Just a few questions about a couple plots, but they are more curiosities.

processing/0_decide_thresholds.py
ajlee21: I can clarify this comment. So a mapping rate of 0 in this case means that a sample had equal mapping to PAO1 and PA14 references since we're taking the difference between mapping rates here. Here I was trying to find samples that specifically map to PAO1

processing/1_create_compendia.py
ajlee21: This is a great comment! I think that looking at this as well as the next point you mentioned will help me to determine if mapping rate is an appropriate metric to use to bin samples.

processing/2_validate_compendia.ipynb
ajlee21: Another great point! My thought was that strain type would be the largest source of variance in the data and so I would have expected clustering based on strain type. But this is not what you see, so looking into other factors like experiment type would be helpful to understand the data more

ajlee21:
Looks like there are fewer PA14 samples with high PA14 mapping, which explains why we see such a reduced number of PA14 binned samples. We may need to used different thresholds for PAO1 and PA14.

image

Looks like the misclassified samples have fairly high mapping rates to PAO1 and PA14 references. They are not just around the threshold set.
image

image

Looks like mapping rate may not be the best method to bin samples. Let's try using median accessory expression instead.

processing/2_validate_compendia.ipynb
ajlee21: There are a lot of study ids so its a little hard to see what is happening, but overall it looks like there is clsutering based on study id. I'm still not clear on why there is not more of a separation in the expression data for strain type, but perhaps this will become more clear later.

image

image

PR #20, opened 2021-04-06
Bin samples using accessory gene expression

ajlee21:
This PR is following up on https://github.com/greenelab/core-accessory-interactome/pull/19 where we binned samples into PAO1 and PA14 compendia based on mapping rates, which didn't seem to distinguish samples sufficiently.

Instead, we will use median accessory gene expression. In our exploratory analysis we found that samples labeled as PAO1 based on SRA annotations had high PAO1 accessory gene expression. Whereas samples labeled as PA14 by SRA had high PA14 accessory gene expression. See plot below where the median expression of PAO1 genes (PAO1 accessory genes) on the x-axis and the median expression of PA14-only genes (PA14 accessory genes) on the y-axis. Each point is a sample.

image

Now binning samples by median accessory gene expression we get:
image

image

As expected, the PAO1 binned samples align along the PAO1-only axis. Similarly, for samples within our binned PA14 compendium, we expect that samples will align along the PA14-axis.

However, we found that our binned samples still include some other SRA-labeled strain types (i.e. Within the PAO1 binned compendium there are samples that SRA labeled as PAK or Clinical). Looking at the distribution of the median accessory gene expression for these non-PAO1 SRA labeled samples (i.e. PAK, Clinical) their expression is very low (blue) compared to all other PAO1 labeled strains (grey).  
image

Similarly for PA14
image

Based on this we increased the threshold applied to the median accessory gene expression. Our final compendia looks like:

image
image

image
image

Also note that our compendium is about ~800 samples for PAO1 and ~500 for PA14, which is close to the numbers that SRA annotates.


nrosed:
Nice! This looks good :)

processing/1_create_compendia.py
```diff
@@ -0,0 +1,249 @@
...
+#
+# Quick check comparing our binned labels compared with SRA annotations
+
+pao1_expression_label["Strain type"].value_counts()
+
+# **Manually check that these PA14 are mislabeled**
+# * Clinical ones can be removed by increasing threshold
+
+pa14_expression_label["Strain type"].value_counts()
+
```
nrosed, line 242: Nice results!

processing/2_validate_compendia.py
```diff
@@ -0,0 +1,561 @@
...
+    plot_title=pn.element_text(family="sans-serif", size=15),
+    axis_text=pn.element_text(family="sans-serif", size=12),
+    axis_title=pn.element_text(family="sans-serif", size=15),
+)
+fig8 += pn.guides(colour=pn.guide_legend(override_aes={"alpha": 1}))
+
+print(fig8)
+# -
+
+# There are a lot of study ids so its a little hard to see what is happening, but overall it looks like there is clsutering based on study id. I'm still not clear on why there is not more of a separation in the expression data for strain type, but perhaps this will become more clear later.
```
nrosed: Yeah, I think it will be clearer as you dig more in the data, it may not effect your downstream analyses anyways.

nrosed (approved):
really nice results, it looks like your strains seem to separate nicely now!

PR #23, opened 2021-04-14
Network validation

ajlee21:
This PR is performing a quick validation of the network modules.

Given a list of genes that are within the same operon or regulon, we want to see if they tend to cluster together compared to random genes.

image
image

* The axis cumulative distribution plots are:
    * x: number of modules that operons/regulons/random genes are contained in
    * y: count
* These distribution plots are summing counts as you move from left to right, so a shift in the curves corresponds to a shift in the distribution (i.e. a curve shifted to the right means that the distribution is shifted to the right)
    * Looking at the red curve, there are ~3000 operon genes that are found in either 0 or 1 modules. If we move over the the right, then there are >3000 operon genes that are found in 0, 1 or 2 modules.
    * Then if you compare the blue and the red curves, the vertical distance between the two curves tells you how much of a shift there is between the distributions.

* We can perform [Kolmogorov-Smirnov test](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test) to compare the distribution of module counts for genes in regulons/operons versus random genes. The KS test will quantify the difference in the cumulative distribution curves.
* Based on the KS test, there is a significant difference between the two distributions as we would expect. Though there is a very large significance for the operons due to the large sample size. There are~3K operons as opposed to the 17 regulons, which has a very modestly significant p-value.  This difference is fairly consistent across thresholds, though as we decrease the threshold used to create the network, the significance for the operons increases slightly but the significance for the regulons decreases slightly.
* This is a quick gut check that the modules seem to make some biological sense before we move forward with more analysis of the modules. Though more tweeks to these modules will likely happen in the future.


ajlee21:
Suggestion by Casey: There are many operons composed of a single gene. Try filtering these out from our analysis

networks/2_module_validation.py
```diff
@@ -0,0 +1,300 @@
...
+)
+cumulative_distribution(
+    pao1_operon["Num_random_modules"],
+    label="Number of modules containing random genes",
+    color="blue",
+)
+_ = plt.legend()
+plt.title("Distribution of module counts (operon vs random genes)")
+
+scipy.stats.ks_2samp(
```
jjc2718, line 255: I'm still a bit confused about exactly what this plot is showing. Is each value on the y-axis an operon? So the value at "1" on the x-axis is the number of operons where the genes in that operon are contained in a single module?

Then the increase at "2" on the x-axis would be the number of operons that are spread across 2 different modules, and so on?

If this is right, it might be helpful to label the axes with something like that (I'm not sure exactly what the labels should be, but I don't think it's super obvious what the axes are here). I guess you have some detail in the "takeaway" section at the bottom, but it may be good to describe it a little on the axes themselves too.

networks/2_module_validation.py
```diff
@@ -0,0 +1,300 @@
...
+    pao1_operon["Num_operon_modules"], pao1_operon["Num_random_modules"]
+)
+# -
+
+fig = sns.displot(
+    pao1_operon_toplot,
+    x="value",
+    hue="variable",
+)
+plt.title("Distribution of module counts (operon vs random genes)")
```
jjc2718, line 265: So is this the same as above, but a PMF (not cumulative) instead of a CDF (cumulative)?

networks/2_module_validation.py
```diff
@@ -0,0 +1,300 @@
...
+#
+# We expect that genes within the same operon or regulon will cluster together (i.e. be within the same module). To test this we will compare the distribution of the number of modules that contain genes within the same regulon vs the number of modules that contain random genes
+
+# +
+# Load PAO1 regulon and operon file
+pao1_regulon_filename = paths.PAO1_REGULON
+pao1_operon_filename = paths.PAO1_OPERON
+
+# Load membership for specific threshold
+corr_threshold = 0.9
```
jjc2718, line 121: Do these plots change much/at all if you vary the threshold?

networks/2_module_validation.py
```diff
@@ -0,0 +1,300 @@
...
+#     * x: number of modules that operons/regulons/random genes are contained in
+#     * y: count
+# * These distribution plots are summing counts as you move from left to right, so a shift in the curves corresponds to a shift in the distribution (i.e. a curve shifted to the right means that the distribution is shifted to the right)
+#
+# * We can perform [Kolmogorov-Smirnov test](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test) to compare the distribution of module counts for genes in regulons/operons versus random genes. The KS test will quantify the difference in the cumulative distribution curves.
+# * Based on the KS test, there is a significant difference between the two distributions as we would expect. Though there is a very large significance for the operons due to the large sample size. There are~3K operons as opposed to the 17 regulons, which has a very modestly significant p-value. This difference is fairly consistent across thresholds, though as we decrease the threshold used to create the network, the significance for the operons increases slightly but the significance for the regulons decreases slightly.
```
jjc2718, line 300: This is what I would expect too, but I wonder how much of this is driven by the single large module. It might be interesting to look at the number of single-module operons and regulons that are in the largest module, as opposed to those where the genes are all contained in a smaller module.

Subjectively I'd find the relationships to be more meaningful if the module is smaller (e.g. if an operon with 5 genes is contained in a module consisting of 10 total genes, this seems more biologically/functionally meaningful than an operon with 5 genes contained in a module consisting of 500 genes).

I guess maybe you could make this quantitative with something like a hypergeometric test, but a simple/straightforward first step might just be looking at the sizes of the modules that contain whole operons or regulons.

jjc2718 (approved):
Looks good! I left some inline questions/comments as usual, but this is definitely an encouraging start.

networks/2_module_validation.py
```diff
@@ -0,0 +1,300 @@
...
+# Load membership for specific threshold
+corr_threshold = 0.9
+pao1_membership_filename = f"pao1_membership_{corr_threshold}.tsv"
+
+pao1_membership = pd.read_csv(pao1_membership_filename, sep="\t", header=0, index_col=0)
+# -
+
+pao1_membership.head()
+
+# ### Format operon/regulon files
```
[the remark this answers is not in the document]
ajlee21: Yes that's right. I'll make sure to add this definition to the notebook too.

So [operons](https://en.wikipedia.org/wiki/Operon#:~:text=An%20operon%20is%20made%20up,transcription%20of%20the%20structural%20genes.) are a group of genes that share a promoter (DNA sequence that is recognized by RNA polymerase and enables transcription) and an operator (DNA sequence that repressor binds to and blocks RNA polymerase). Therefore these group of genes are transcribed or turned off together (so we would expect a very high correlation amongst these genes)

[Regulons](https://en.wikipedia.org/wiki/Regulon) are a group of genes that are regulated by the same regulatory protein. A regulon can be composed of multiple operons.

networks/2_module_validation.py
```diff
@@ -0,0 +1,300 @@
...
+        fig.set_title(
+            f"Histogram of size of modules using threshold {threshold_list[i]}",
+            fontsize=12,
+        )
+        handles, labels = fig.get_legend_handles_labels()
+        fig.legend(handles, labels, loc="center right")
+
+
+# Plot distribution of modules per threshold
+plot_dist_modules(corr_threshold_list)
```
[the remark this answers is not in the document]
ajlee21: Good eye. I didn't notice that before. Something I'll have to think about why/if that makes sense.

networks/2_module_validation.py
ajlee21: Yes these are a couple different visualizations of the same data. I'll update the titles to be more specific

networks/2_module_validation.py
ajlee21: I'm also still trying to wrap my head around these plots, let's see if this is more clear:

y-axis = the cumulative count of genes within operon/regulon (red) or random genes (blue).
x-axis = The number of modules that genes are contained in

Looking at the operon plot, the value at "1" on the x-axis says that there are ~3000 operon genes found in exactly 1 module, and there are ~2700 random genes found in exactly 1 module

Then the increase at "2" on the x-axis is the number of operons or random genes that are spread across 2 or 1 different modules (this is the cumulative part). There are >3000 operon genes that are found in either 1 or 2 modules. There are > 3000 random genes that are found in either 1 or 2 modules.

Then if you compare the blue and the red curves, the vertical distance between the two curves tells you how much of a shift there is between the distributions.

I will definitely add axis labels here to help. Please let me know if this explanation helps or if its still confusing.

networks/2_module_validation.py
ajlee21: This trend held across varying thresholds: as we decreased thresholds, the KS stat for the operons became more significant while the KS stat for the regulons became slightly less signficant (0.08). The visual trend held despite differences in the p-value thought is is probably more due to the sample size differences since there are only 17 regulons compared to 3K operons

networks/2_module_validation.py
ajlee21: This is a very good point. I will look at the size of the modules. Another thing I can try is to just remove the modules with >1K genes and see what the trends look like too.

networks/2_module_validation.py
jjc2718: Ok, this helps!

Just to clarify one small point: when you say "random genes spread across 2 modules", you really mean "random gene sets (of equal size to an operon) spread across 2 modules", right? I don't think it would make sense for a single gene to be spread across multiple modules - each gene should be in exactly one module as far as I know, unless you're finding overlapping modules (but I don't think this should be the case with most hierarchical clustering implementations).

networks/2_module_validation.py
ajlee21: AFTER removing modules with > 1000 genes
* Using the KS test, there is a significant difference (across thresholds) between the operon and random distribution, as we would expect.
* There is only a significant difference between the regulon and random distributions at lower thresholds (0.7 and below). The lack of significance at higher thresholds is likely due to the small sample size (i.e. 6 or 10 regulons have genes that are contained in modules with fewer than 1000 genes)

networks/2_module_validation.py
ajlee21: Yes that is correct - a random set (size matched with the operon) where the genes in that set are found in 2 modules

ajlee21: Sorry slight error on my part, y-axis is the cumulative number of operons/regulons/random groups NOT genes. So your initial statement was correct. Its the number of Y operons spread across X modules in the network. I'll need to update this text and labels

PR #25, opened 2021-04-29
Apply correlation correction

ajlee21:
Previously we noticed that when we apply clustering on gene expression correlation matrix, gene pairs tended to cluster into a single large module. This observation is consistent with a [previous study](https://link.springer.com/article/10.1186/1471-2164-7-187), which found that KEGG (a database that containes genes or proteins annotated with specific biological processes as reported in the literature) is bias in some biological processes represented. Figure 1C demonstrates that a large fraction of gene pairs are ribosomal relationships - in the top 0.1% most co-expressed genes, 99% belong to the ribosome pathway. Furthermore, protein function prediction based on co-expression drop dramatically after removing the ribisome pathway (Figure 1A, B). This finding is consistent with our observation when we calculate the correlation of the raw gene expression data.
We found one large highly correlated module that is likely driven by genes related to a single biological process.

**Challenge:** This very dominant global signal can mask more specific signals in the data.
In this PR we tried several approaches to extracting correlations between genes that correct for this:
1. Transformed raw data and then apply correlation
2. Apply dimensional reduction using PCA or SVD (SPELL) on raw data and then apply correlation
2. Scale the high-degreeness in the correlation matrix (SEEK/Hetio) 

Main changes:
1. Performed exploration of different correlation correction methods listed above. The notebooks can be found in `archive` and the results can be found here: https://docs.google.com/presentation/d/1mLdLk6j3C-XyoxsKvJ0Db7_cbNlMZi7eaNrCH6O8LcM/edit#slide=id.p

2. Final correlation correction analysis using SPELL method: `1_correlation_analysis.ipynb`
 
There is a lot happening in (1) but most of the code is redundant to what is seen in (2) and I wanted to keep the (1) notebooks for reference but shouldn't need much of a review. I would focus on the changes in (2)


networks/1_correlation_analysis.py
```diff
@@ -59,143 +59,177 @@
...
+    paths.LOCAL_DATA_DIR, "pao1_log_spell_clustermap.png"
 )
-h1.savefig(pao1_clustermap_filename, dpi=300)
+h1a.savefig(pao1_log_spell_filename, dpi=300)
 
 # +
-plt.figure(figsize=(20, 20))
-h2 = sns.clustermap(pa14_corr.abs(), cmap="viridis")
-h2.fig.suptitle(f"Correlation of PA14 genes using threshold={corr_threshold}")
+h2a = sns.clustermap(pa14_corr_log_spell.abs(), cmap="viridis", figsize=(20, 20))
```
jjc2718, line 214: It's interesting that the log transform really seems to be a necessary step in getting rid of the block diagonal structure (i.e. the large, highly connected network components).

Do you think the large communities were due to correlations in genes with very low expression? So the log transformation may have smoothed out these low expression values? (Just a guess)

networks/archive/1a_transformation_correlation_analysis.py
```diff
@@ -0,0 +1,135 @@
...
+plt.figure(figsize=(20, 20))
+h3 = sns.clustermap(pao1_corr_normalized.abs(), cmap="viridis")
+h3.fig.suptitle("Correlation of 0-1 normalized PAO1 genes", y=1.05)
+
+plt.figure(figsize=(20, 20))
+h4 = sns.clustermap(pa14_corr_normalized.abs(), cmap="viridis")
+h4.fig.suptitle("Correlation of 0-1 normalized PA14 genes", y=1.05)
+
+# **Takeaway:**
+# Here we are using the Pearson correlation metric, which is defined by the formula:
```
jjc2718, line 128: I can't remember - have you considered trying [Spearman rank correlation](https://en.wikipedia.org/wiki/Spearman%27s_rank_correlation_coefficient) instead of Pearson? That could help in the cases where you have varying scales/ranges (Spearman is essentially the Pearson correlation of the rank vectors).

I think people sometimes use Spearman to construct co-expression networks (see, e.g. [this paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3511551/)).

Not that you necessarily need more things to try - seems like the log-transform + SPELL approach is giving you what you want, so feel free to stick with that if you want to.

networks/archive/1b_corrected_correlation_analysis.py
```diff
@@ -0,0 +1,506 @@
...
+h8 = sns.clustermap(pa14_corr_hetio_df.abs(), cmap="viridis")
+h8.fig.suptitle("Correlation of PA14 genes using (Hetio corrected)", y=1.05)"""
+
+# ## Subtract the mean
+#
+# Inspired by hetio search, here we are subtracting the mean of the row and column from each correlation value.
+#
+# _implementation note:_
+#
+# We are subtracting the row mean and column mean separately based on a simple proof where $A_ij$ is the value in row _i_ and column _j_, There are a total of _N_ columns and rows since the matrix is square.
```
jjc2718, line 455: Yay, math!

networks/archive/1b_corrected_correlation_analysis.py
```diff
@@ -0,0 +1,506 @@
...
+plt.ylabel("Variance ratio")
+plt.xlabel("Principal components")
+plt.show()
+# -
+
+# Correlation
+pao1_corr_pca = pao1_pc_weights_df.corr()
+pa14_corr_pca = pa14_pc_weights_df.corr()
+
+"""# Plot heatmap
```
jjc2718, line 277: Do you want to plot these?

networks/archive/1b_corrected_correlation_analysis.py
```diff
@@ -0,0 +1,506 @@
...
+
+"""plt.figure(figsize=(20, 20))
+h6 = sns.clustermap(pa14_corr_seek.abs(), cmap="viridis")
+h6.fig.suptitle("Correlation of PA14 genes (SEEK corrected)", y=1.05)"""
+
+# ## Hetio
+#
+# https://het.io/ allows users to search a heterogeneous network (i.e. are graphs with multiple node and edge types like disease or gene nodes) to find connections between entities such as eye and breast cancer. It is used as a hypothesis generating tool.
+#
+# To search the network, they empoloy a correction for hubbiness. Here we use their correction. This correction is similar to SEEK in that it is applying a correction using both the source and target node to deal with the asymmetry that we encountered.
```
jjc2718, line 395: Interesting that this didn't seem to work for you!

Although maybe your problem isn't really specifically "hubbiness", since you're not looking at network paths like they were. Still, I might have expected it to help a bit with correcting for highly connected parts of the network.

networks/1_correlation_analysis.py
```diff
@@ -16,35 +16,35 @@
...
 
-# +
 # Params
-corr_threshold = 0.5
-
-# Output files
-pao1_membership_filename = f"pao1_membership_{corr_threshold}.tsv"
-pa14_membership_filename = f"pa14_membership_{corr_threshold}.tsv"
-# -
+num_singular_values = 300
```
jjc2718, line 46: Did you try varying these numbers at all?

Would it be possible to make variance explained plots like people do for PCA, to see how much of the variance in the data you're picking up for various numbers of singular values? It seems like you can do this if you use the [sklearn version of SVD](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.TruncatedSVD.html), although I'm not sure if the numpy implementation provides this function out of the box.

jjc2718 (approved):
Looks good to me! Just a few comments.

Glad to see that you were able to remove the obvious block structure from the correlation matrices. Are you planning to rerun the clustering analyses with the corrected correlations next, or is there something else you're going to do?

networks/0_examine_data.py
```diff
@@ -0,0 +1,107 @@
...
+#
+# These pair plots also give us a rough sense for how correlated genes are. We would expect co-operonic genes to be more highly correlated compared to a random set of genes, which we do see. Some correlations appear not as strong due to the differences in scales between genes - something to consider when we are looking at correlations between genes.
+
+# ## Plot distribution of pairwise distances
+#
+# This will particularly help to inform the parameters we use for DBSCAN, which is density based. Here we looking at the distribution of both global distances and local distances. Global distances are defined using `pdist`, which takes the pairwise Euclidean distance of each of the correlation vectors (so the distance between gene `p` and gene `q` is based on the difference in correlation between `p` and all other genes, and `q` and all other genes). Whereas the local distance is defined as 1 - |correlation(`p`, `q`)|
+
+# Get distribution of pairwise distances to determine a cutoff defining what a dense region should be
+f1 = sns.displot(pdist(pao1_corr))
```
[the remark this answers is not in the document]
ajlee21: Yup! This was originally part of the correlation notebook and I moved it outside, but I think it actually should be in the correlation notebook. Good catch

networks/1_correlation_analysis.py
```diff
@@ -16,35 +16,35 @@
...
+# Furthermore, protein function prediction based on co-expression drop dramatically after removing the ribisome pathway (Figure 1A, B).
+#
+# This notebook applies different corrections to try to remove this very dominant global signal in the data. This notebook follows from [1a_transformation_correlation_analysis.ipynb](1a_transformation_correlation_analysis.ipynb). Here we are applying dimensionality reduction techniques in addition to scaling the data.
```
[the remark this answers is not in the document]
ajlee21: Good point. I wasn't keeping track of the numbering because these were archived. These notebooks are independent so I think I will make a note and remove the numbering here

networks/archive/1b_corrected_correlation_analysis.py
ajlee21: Since it takes 2-3 minutes per correlation plot and I was playing around with different variables, I was commenting out different plots. So yes, this was intended

networks/1_correlation_analysis.py
ajlee21: So based on the math, I believe these large communities are due to those genes that are in the right tail of the distribution - these are genes that will cause the covariance to be very high just because of how far it deviates from the mean.

networks/1_correlation_analysis.py
ajlee21: Yes I varied this number a bit, but 300 seemed reasonable given that eADAGE uses 300, though that is a nonlinear model.

So for both SVD and PCA, the variance explained was captured in fewer than 10 PCs and 20 singular vectors, but with such few singular vectors or PCs, there we were still seeing very large modules.

ajlee21: These curves should be in [1b_corrected_correlation_analysis.ipynb](networks/archive/1b_corrected_correlation_analysis.ipynb)

networks/archive/1b_corrected_correlation_analysis.py
ajlee21: These curves should also be in [1b_corrected_correlation_analysis.ipynb](networks/archive/1b_corrected_correlation_analysis.ipynb)

networks/archive/1a_transformation_correlation_analysis.py
ajlee21: I did try Spearman correlation earlier. See [slide 10](https://docs.google.com/presentation/d/10Pr_oD5jL8GxCHLg1Qb16A6br31E2COgZIThjnj5ZyI/edit#slide=id.gd398a77b8a_0_4).

It does look like Spearman can help a bit, but I don't think results should change too much if we switch.

PR #26, opened 2021-05-19
Update threshold for compendia and reprocess

ajlee21:
This PR is updating the threshold used to partition Pseudomonas gene expression data into PAO1 and PA14 compendia.

This threshold is based on the distribution of median accessory gene expression (i.e. we expect PAO1 samples to have high PAO1 median accessory gene expression and 0 PA14 median accessory expression). To determine the threshold we will look at the distribution of median accessory gene expression in samples SRA labeled as PAO1 vs non-PAO1.

The main new change is found in `0_decide_threshold.ipynb` that explores where to place the threshold to separate the PAO1 (grey) vs non-PAO1 (blue) samples.

image

For separating between PA14 vs non-PA14 samples:
image

More details can be found in the readme


processing/0_decide_threshold.py
```diff
@@ -0,0 +1,222 @@
...
+    "median_acc_expression_pao1",
+]
+
+non_pao1_sra = pao1_pa14_acc_expression.loc[
+    pao1_pa14_acc_expression["Strain type_pao1"] != "PAO1",
+    "median_acc_expression_pao1",
+]
+# -
+
+pao1_threshold = 5
```
nrosed, line 196: How did you decide 5?

nrosed: probably overkill.... but could you use a classifier here?

processing/0_decide_threshold.py
```diff
@@ -0,0 +1,222 @@
...
+    "median_acc_expression_pa14",
+]
+
+non_pa14_sra = pao1_pa14_acc_expression.loc[
+    pao1_pa14_acc_expression["Strain type_pa14"] != "PA14",
+    "median_acc_expression_pa14",
+]
+# -
+
+pa14_threshold = 5
```
nrosed, line 215: same question as above for the pao1_threshold definition

nrosed (approved):
Looks good! Just one question added in the notebook about the selection of 5 as the cutoff.

processing/0_decide_threshold.py
ajlee21: Our goal here is to define a threshold to determine if a sample if PAO1 or not. We used known labels from SRA to do this. Specifically, we found a threshold that separated between the distribution of PAO1 samples (grey) vs non-PAO1 samples (blue). 

I will clarify this in the notebook.

PR #27, opened 2021-06-01
Update clustering scripts

ajlee21:
This PR updates the clustering scripts, which takes the correlation matrix (based on how similar gene expression profiles are) and clusters genes into modules.

The current data being used here is not the final version and is currently being updated. So this PR is updating the clustering scripts in preparation for the new version of data and therefore some of the parameters will likely need to be tweeked.


networks/1_correlation_analysis.py
```diff
@@ -265,13 +265,13 @@
...
-"""pao1_log_spell_mat_filename = os.path.join(
+pao1_log_spell_mat_filename = os.path.join(
     paths.LOCAL_DATA_DIR, "pao1_log_spell_mat.tsv"
 )
 pa14_log_spell_mat_filename = os.path.join(
     paths.LOCAL_DATA_DIR, "pa14_log_spell_mat.tsv"
 )
 pao1_corr_log_spell.to_csv(pao1_log_spell_mat_filename, sep="\t")
-pa14_corr_log_spell.to_csv(pa14_log_spell_mat_filename, sep="\t")"""
+pa14_corr_log_spell.to_csv(pa14_log_spell_mat_filename, sep="\t")
```
miltondp, line 275: This was uncommented, right? I mean, it was as a string before?

networks/2_get_network_communities.py
```diff
@@ -36,19 +36,12 @@
...
-)
-pa14_corr_filename = os.path.join(
-    paths.LOCAL_DATA_DIR, f"pa14_corr_{corr_threshold}.tsv"
-)
-
-# +
-## TO DO
-# Create variable to determine which clustering method to use
+pao1_corr_filename = os.path.join(paths.LOCAL_DATA_DIR, "pao1_log_spell_mat.tsv")
+pa14_corr_filename = os.path.join(paths.LOCAL_DATA_DIR, "pa14_log_spell_mat.tsv")
```
miltondp, line 44: Minor: you might want to switch to pathlib in the future if you consider it more clear.

networks/2_get_network_communities.py
```diff
@@ -64,40 +57,44 @@
 # A cluster is a set of core samples that can be built by recursively taking a core sample, finding all of its neighbors that are core samples, finding all of their neighbors that are core samples, and so on. A cluster also has a set of non-core samples, which are samples that are neighbors of a core sample in the cluster but are not themselves core samples. Intuitively, these samples are on the fringes of a cluster.
 #
 # * We define a core sample as being a sample in the dataset such that there exist `min_samples` other samples within a distance of `eps`, which are defined as neighbors of the core sample.
-# * Using all default values
+# * Here we use `eps=8` based on the observations in the [prevous notebook](1_correlation_analysis.ipynb)
```
miltondp, line 60: I'm not sure if I found these analyses/conclusions in `1_correlation_analysis.ipynb`. Are you inferring this value (`eps=8`) from the distribution plots?

networks/2_get_network_communities.py
```diff
@@ -64,40 +57,44 @@
...
 #
 
-# +
 # Clustering using DBSCAN
-# pao1_clustering = DBSCAN().fit(pao1_corr)
-# pa14_clustering = DBSCAN().fit(pa14_corr)
-# -
+if cluster_method == "dbscan":
+    pao1_clustering = DBSCAN(eps=8).fit(pao1_corr)
+    pa14_clustering = DBSCAN(eps=8).fit(pa14_corr)
```
miltondp, line 66: This is a minor and very personal style comment: I prefer to move all settings/parameters (like `eps` here or `n_clusters` and `distance_threshold` for hierarchical clustering) to the top of the notebook, like in your "Set user parameters" section. This way all parameters that can be changed are in one place and are not lost in the middle of the code.

miltondp (approved):
Looks good to me, I left some minor comments and questions. Looks like most of the code was already reviewed. Let me know please if you wanted me to focus on something in particular.

networks/1_correlation_analysis.py
ajlee21: Yes I forgot to uncomment this out from the previous PR

networks/2_get_network_communities.py
ajlee21: Nice catch. I'll add these

networks/2_get_network_communities.py
ajlee21: Yes I'm using the distribution plots for pdist to determine this `eps` value. I will update the comment to make this more clear though

ajlee21:
Thank you for your suggestions @miltondp if you have any additional thoughts on what method to use for clustering moving forward that would be helpful. For now I am applying several

miltondp:
Spectral clustering is a nice method that can find more complex structures. It is similar to DBSCAN, but uses the spectrum of the Laplacian matrix (if I recall correctly), and then applies k-means on it using the k top eigenvectors.

miltondp:
Yet another "meta-method" could be to use all other methods, and then combine the clustering results into a single solution. You can also estimate the number of clusters with this approach. It's called "consensus clustering" or "cluster ensembles", and it is an ensemble approach similar to classification.

miltondp:
Of course, I would try simpler approaches first and see if that answers your question. If you need more complexity, these other approaches could be helpful.

PR #28, opened 2021-06-04
Update validation scripts

ajlee21:
This PR has the following major changes:

1. Updating the existing validation analysis to use the new clustering approach compared to using thresholding of correlations scores (`3_module_validation.ipynb`)
2. Adds notebooks to compare the composition of modules found using the array data vs RNA-seq data. `0_process_array_data.ipynb` processes the array and rna-seq compendia to use the same set of genes. Then we re-run the correlation and clustering notebooks to get modules for the 2 compendia and `3a_validate_array_vs_rnaseq_modules.ipynb` compares the module composition

This PR is preparing the validation analysis scripts that will need to be run once we get the final version of the input data. Based on the results of the validation analysis using our current version of the data, there looks like we'll need to play around with the clustering parameters and dig a little into why the array vs rna-seq modules are not more consistent.


networks/3a_validate_array_vs_rnaseq_modules.py
```diff
@@ -0,0 +1,217 @@
...
+pao1_rnaseq_corr_log_spell = rnaseq_U_df.iloc[:, :num_singular_values_log].T.corr()
+
+# Note: Here we plot the heatmaps to verify that the correlation of log + SPELL transformed data looks as expected (i.e. there is not a single large cluster)
+
+# Plot heatmap
+"""h1a = sns.clustermap(pao1_array_corr_log_spell.abs(), cmap="viridis", figsize=(20, 20))
+h1a.fig.suptitle(
+    f"log transform + SPELL corrected using {num_singular_values_log} vectors (PAO1 array)",
+    y=1.05,
+)"""
```
miltondp, line 114: Just curious: why not comment the code instead of making it a string?

networks/3a_validate_array_vs_rnaseq_modules.py
```diff
@@ -0,0 +1,217 @@
...
+    ).fit(pao1_rnaseq_corr_log_spell)
+
+# Clustering using affinity propogation
+if cluster_method == "affinity":
+    pao1_array_clustering = AffinityPropagation(random_state=0).fit(
+        pao1_array_corr_log_spell
+    )
+    pao1_rnaseq_clustering = AffinityPropagation(
+        random_state=1, damping=affinity_damping
+    ).fit(pao1_rnaseq_corr_log_spell)
```
miltondp, line 148: Here `pao1_array_corr_log_spell`, the input to all algorithms, is a squared matrix of correlations, right? I see that `AffinityPropagation` supports features or affinities as input, but you have to provide the `affinity='precomputed'` parameter. Not sure if I'm missing something.

networks/3a_validate_array_vs_rnaseq_modules.py
```diff
@@ -0,0 +1,217 @@
...
+
+# +
+# Get module membership for a single threshold
+# Format and save output to have columns: gene_id | group_id
+pao1_array_membership_df = pd.DataFrame(
+    data={"module id": pao1_array_clustering.labels_},
+    index=pao1_array_corr_log_spell.index,
```
miltondp, line 155: Minor: I always prefer to use a copy of the `index` of other dataframes, since I remember having some weird issues using directly the original instances.

networks/3a_validate_array_vs_rnaseq_modules.py
miltondp: Another question: why are you using a different set of parameters here for RNA-seq and array data?

networks/3a_validate_array_vs_rnaseq_modules.py
```diff
@@ -0,0 +1,217 @@
...
+
+print(len(consistent_array_modules) / len(array_module_mapping))
+
+# +
+rnaseq_module_mapping = {}
+for grp_id, grp in pao1_rnaseq_membership_df.groupby("module id"):
+    grp_mapped = pao1_array_membership_df.loc[grp.index]
+    rnaseq_module_mapping[grp_id] = list(grp_mapped["module id"].unique())
+
+rnaseq_module_mapping
```
miltondp, line 202: I'm reading aloud this piece of code to see if I understand: for rna-seq data, `grp_id` is the cluster id, `grp` contains the list of genes in this cluster, and when you do `grp.index` you are getting that list of genes in the cluster. So `rnaseq_module_mapping[grp_id]` contains the list of cluster ids in array data that contain each of these genes in the cluster `grp_id` in rnaq-seq data.

miltondp: So what you would like to find is that each single rna-seq cluster map to a single array cluster, right? This is what you call "consistency"?

networks/3a_validate_array_vs_rnaseq_modules.py
```diff
@@ -0,0 +1,217 @@
...
+
+# +
+# Are there any rnaseq modules that map to a single array module?
+consistent_rnaseq_modules = []
+for rnaseq_module_id, list_array_module_ids in rnaseq_module_mapping.items():
+    if len(list_array_module_ids) == 1:
+        print(rnaseq_module_id, list_array_module_ids)
+        consistent_rnaseq_modules.append(rnaseq_module_id)
+
+print(len(consistent_rnaseq_modules) / len(rnaseq_module_mapping))
```
miltondp, line 212: If I'm not missing something, I think you could use more standard ways to compute the similarity between two clustering partitions, like the adjusted Rand index (ARI, it's available in sklearn). If each rna-seq cluster map to a single array cluster, then you get an ARI=1.0
I know that here you are doing it manually to have a percentage/more interpretable measure, but I wonder what's the ARI value of these two data partitions. Also, you have [other clustering measures](https://scikit-learn.org/stable/modules/classes.html#sklearn-metrics-metrics) (under "Clustering metrics") that let you know, for instance, if clusters in one partition are "contained" in other clusters, what could be interesting for your question I think (like the homogeneity/completeness scores).

miltondp: I mention this because it could happen that there is not a perfect match between clusters in both data types, but maybe in general gene pairs are clustered in the same clusters. You will miss this if you don't use a pair-counting-based measure like ARI.

networks/3a_validate_array_vs_rnaseq_modules.py
```diff
@@ -0,0 +1,217 @@
...
+# -
+
+# **Observation:**
+#
+# * Only ~3% of modules are consistent using RNA-seq and array data. We would have expected more of an overlap. What is causing this?
```
miltondp, line 217: The first thing I would do is to use exactly the same parameter set for both data types. I think you are using a different dumping factor for Affinity propagation. Not sure, though, if you already analyzed this and have a reason for doing it.

The second thing is that you are running the algorithm only once if I'm not mistaken. I'm not very familiar with affinity propagation, but the fact that it supports a `random_state` parameter makes me think that if you run it several times, it will produce different partitions.

Third, what's the result with the other clustering algorithms? Both of them, hierarchical clustering and DBSCAN, have the advantage of being deterministic, so it would be interesting to see how they performed. Other algorithms like Spectral clustering allow you to specify `n_init`, such as 100 or 200, meaning that the algorithm will be run 100 or 200 times and then it will return the "best" partition. That would be probably a better strategy than just taking a random run (such as with Affinity propagation).

A fourth point would be to report also the similarity given by measures like ARI.

miltondp (approved):
Alex, I left you some comments, questions, and suggestions. Many are minor things, others are ideas to explore. Let me know if I can help with anything.

networks/0_process_array_data.py
```diff
@@ -0,0 +1,101 @@
...
+]
+
+pao1_sample_ids = array_metadata[
+    array_metadata["strain"].apply(
+        lambda strain_name: strain_name in pao1_strain_values
+    )
+]["ml_data_source"]
+
+# Drop any sample ids that are na
+pao1_sample_ids.dropna(inplace=True)
```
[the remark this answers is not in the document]
ajlee21: I _think_ the reason might be just in case you lose track of changes you've made to a dataframe that you've over-written. I usually just set a new variable, but it depends on what I'm doing.

networks/3_module_validation.py
```diff
@@ -24,59 +24,74 @@
...
 np.random.seed(1)
 
 # +
 # User params
+# Params to examine module size
+clustering_method_list = ["dbscan", "hierarchal", "affinity"]
+
+# Params for regulon/operon coverage
+# Clustering method to examine regulon/operon coverage
+method_toexamine = "dbscan"
```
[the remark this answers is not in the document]
ajlee21: yes, I'll update the comment to mention this.

networks/0_process_array_data.py
```diff
@@ -0,0 +1,101 @@
...
+print(pao1_rnaseq_compendium.shape)
+pao1_rnaseq_compendium.head()
+
+# +
+# Get shared genes
+pao1_array_gene_ids = pao1_array_compendium.columns
+pao1_rnaseq_gene_ids = pao1_rnaseq_compendium.columns
+
+shared_gene_ids = list(set(pao1_array_gene_ids).intersection(set(pao1_rnaseq_gene_ids)))
+print(len(shared_gene_ids))
```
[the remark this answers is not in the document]
ajlee21: yup

networks/3_module_validation.py
```diff
@@ -24,59 +24,74 @@
...
 
-# +
-corr_threshold_list = [0.9, 0.8, 0.7, 0.6, 0.5]
-
-for corr_threshold in corr_threshold_list:
-    print(f"Modules using correlation threshold: {corr_threshold}")
-    pao1_membership_filename = f"pao1_membership_{corr_threshold}.tsv"
-    pa14_membership_filename = f"pa14_membership_{corr_threshold}.tsv"
+for method_name in clustering_method_list:
+    print(f"Modules using correlation threshold: {method_name}")
```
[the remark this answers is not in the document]
ajlee21: Good catch. I will update this

networks/3_module_validation.py
```diff
@@ -102,23 +117,24 @@ def plot_dist_modules(threshold_list):
...
 
 
 # Plot distribution of modules per threshold
-plot_dist_modules(corr_threshold_list)
+plot_dist_modules(clustering_method_list)
 
 # **Takeaway:**
-# * Looks like as we decrease our correlation threshold (i.e. connections don't need to be very strong between nodes), more modules are able to form which is what I would expect.
-# * Genes in PA14 are less concentrated in one large module - there are many more small modules found.
+# * Looks like there is one large modules using DBSCAN and hierarachal clustering
```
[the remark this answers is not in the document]
ajlee21: This module should be filtered out https://github.com/greenelab/core-accessory-interactome/blob/0996cf12b6f1833ea9b4ae30efdd605610fd0914/networks/3_module_validation.py#L148

networks/3a_validate_array_vs_rnaseq_modules.py
ajlee21: This was an easy way to comment multiple lines

networks/3_module_validation.py
```diff
@@ -102,23 +117,24 @@ def plot_dist_modules(threshold_list):
...
+# * There are more even sized modules using affinity propogation
```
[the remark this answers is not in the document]
ajlee21: So each distribution should show the distribution of the size of modules where the x-axis is the module id and the height of the bar is the number of genes contained within the module, but since this is a histogram we have bins which are grouping several modules together. So it would be more clear if we had each module plotted separately.

networks/3a_validate_array_vs_rnaseq_modules.py
ajlee21: Yes your interpretation is correct. I'll try to add more clarify to this.

networks/3a_validate_array_vs_rnaseq_modules.py
ajlee21: Good point. I'll need to think a bit more about the parameter settings I'm using. The reason for the difference is because the model didn't converge unless i increased the damping factor here.

networks/3a_validate_array_vs_rnaseq_modules.py
ajlee21: Very good point. I'll look into this.

networks/3a_validate_array_vs_rnaseq_modules.py
ajlee21: Thank you! I really appreciate all these very helpful feedback. I'll definitely update my clustering and validation steps accordingly moving forward.

PR #29, opened 2021-06-14
Add community detection and random walk approaches

ajlee21:
This PR is testing out a couple of additional approaches to identify modules of co-acting genes: community detection and random walk.

Up until this point, we have been applying clustering on our correlation matrix to identify groups of co-acting genes. Here we are testing out community detection and random walk approaches to see if applying this module detection on the network data as opposed clustering on the nodes helps. We also played with the user parameters a bit.

Changes include:
1. Adding `2b_community_detection.ipynb` notebook to perform community detection,
2. Updated `3_module_validation.ipynb` notebook to plot validation results for the community detection results.

Our expectation on size of modules would be 2-50 genes. Most operons have fewer than 10 genes and most regulons have fewer than 100 genes. Based on the results here, it looks like affinity propagation returns results sensible results. We will move forward and calculate composition of modules generated by affinity propagation

In the future we will need to think about the following how to get robust modules:
I can see 2 possible options for this:
1. Generate multiple partitions using a chose community detection method. Then aggregate across those partitions to get a single "consensus partitions". Then I can calculate the composition of modules based on that consensus partition
2. OR I can get 1 partition from my community detection method and calculate the composition of modules based on that 1 partition. Then I can repeat this process to get multiple compositions per partition.

Right now I'm thinking that 2 is more straightforward to implement though I'll have to figure out how I'll summarize the composition calculations. I'm honestly not sure what aggregation methods exist for 1.


networks/2b_community_detection.py
```diff
@@ -0,0 +1,202 @@
...
+
+# make sure vertex/edge properties exist
+print(pao1_G.es["weight"][:5])
+
+# ## Community detection
+
+# ### Fast-greedy
+# This algorithm starts from a completely unclustered set of nodes and iteratively adds communities such that the modularity (score maximizing within edges and minimizing between edges) is maximized until no additional improvement can be made.
+#
+# Note: Looks like fast-greedy requires a simple graph (i.e. no multiple edges per node), so we use [simplify](https://igraph.org/python/doc/api/igraph._igraph.GraphBase.html#simplify) to combine edges
```
miltondp, line 100: Hm, interesting. So one node/gene is connected only to a single node? This is what "simplify" does?

networks/2b_community_detection.py
```diff
@@ -0,0 +1,202 @@
...
+        pao1_G.simplify()
+    if not pa14_G.is_simple():
+        pa14_G.simplify()
+
+    assert pao1_G.is_simple()
+    assert pa14_G.is_simple()
+
+    # Detection method
+    pao1_partition = pao1_G.community_fastgreedy(weights=pao1_G.es["weight"])
+    pa14_partition = pa14_G.community_fastgreedy(weights=pa14_G.es["weight"])
```
miltondp, line 117: Is the partition for these methods a 1d array with integers like the _labels property of sklearn's clustering algorithms?

networks/2b_community_detection.py
```diff
@@ -0,0 +1,202 @@
...
+
+# +
+# pao1_G.vs.attribute_names()
+# -
+
+# ### Walktrap
+# This algorithm performs random walks using a specified step size. Where densely connected areas occur, the random walk becomes “trapped” in local regions that then define communities
```
miltondp, line 124: I liked this simple and very easy-to-follow description. I never used these randomwalk algorithms.

networks/2b_community_detection.py
```diff
@@ -0,0 +1,202 @@
...
+#       jupytext_version: 1.9.1+dev
+#   kernelspec:
+#     display_name: Python [conda env:core_acc] *
+#     language: python
+#     name: conda-env-core_acc-py
+# ---
+
+# # Network community detection
+#
+# This notebook performs community detection approaches to identify network modules.
```
miltondp, line 19: What is basically the difference between applying clustering and community detection algorithms? I never used the latter, and I wonder what is the main difference.

networks/2b_community_detection.py
```diff
@@ -0,0 +1,202 @@
...
+        return membership_df
+
+
+# +
+# pao1_partition.es.attribute_names()
+# -
+
+pao1_membership_df = graph_partition_to_df(pao1_G, pao1_partition, method)
+print(len(pao1_membership_df["module id"].unique()))
+pao1_membership_df.sort_values(by="degree", ascending=False).head()
```
miltondp, line 187: I think responds to a previous question.

miltondp: Here "module id" can be read as the label, right?

networks/3_module_validation.py
```diff
@@ -113,9 +123,14 @@ def plot_dist_modules(clustering_method_list):
...
-# * Looks like there is one large modules using DBSCAN and hierarachal clustering
-# * There are more even sized modules using affinity propogation
-# * For all methods, the size of most modules are 10-50's (Note: as a results of the binning which is summing counts across multiple modules, the height of the binds are in the hundreds)
+# Our expectation on size of modules would be 2-50 genes. Most operons have fewer than 10 genes and most regulons have fewer than 100 genes. Some examples that demonstrate the size of co-expression networks can be found in papers using ADAGE signatures to define modules:
+# * Figure 5 in [eADAGE paper](https://bmcbioinformatics.biomedcentral.com/track/pdf/10.1186/s12859-017-1905-4.pdf)
+# * Figure 7 in [Harty et al. paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6531624/)
+# * Figure 2 in [Doing et al. paper](https://journals.plos.org/plosgenetics/article?id=10.1371/journal.pgen.1008783)
+#
+# What did we find? Which method follows our expectation? This is the one we will move forward.
+# * Looks like there is one large modules using DBSCAN clustering
```
miltondp, line 132: Did you try to vary the parameters of DBSCAN (eps and min_samples) and see if you find module sizes as you expect? Maybe that could be another way of starting these analyses: if you already know what module sizes should be, then for each algorithm you try you could change parameters until finding something you expect.

miltondp (approved):
Looks good to me, Alex! I left some minor comments and questions. Also I paste below some of our conversation in Slack plus some additions/changes:

I believe that option 2 is the best as a start. Keep things as simple as possible, but not simpler (when this happens you could move to option 1). I can give you the code to compute a consensus partition if you want. The nice thing about the consensus approach is that if those modules have complex structures, then you can use different algorithms with different assumptions, and then merge all and get the "average", or get together the gene pairs that tend to be clustered together across all algorithms. Sometimes the final consensus partition could be very different to individual members of the ensemble, and get final structures that are impossible to obtain with simpler algorithms.

Option 1 (consensus) is more useful when you have some diversity in your ensemble; if partitions are all the same, then I would just go with 2.
If you have the labels with core/accessory for all your data, you can just compute the ARI/NMI/etc between the labels and the output of the algorithm. That will give you a number between 0 (no agreement with labels) and 1 (perfect agreement). You can run that multiple times and average the ARI across.

networks/2b_community_detection.py
ajlee21: This was my quick read on `simplify` though I was still not able to get fast-greedy to run. After spending some time to try to debug this, I just moved forward with the other methods since affinity seems to be giving us close to expected results

networks/2b_community_detection.py
ajlee21: Good question! This was my same question as well when Casey mentioned I might want to take a look at community detection methods. So the main difference seems to be that clustering thinks about each gene as a point/entity and performs clustering to place similar genes (i.e. genes that are close together in correlation space) together where each gene belongs to a single module. Whereas, community detection considers genes to be in a network/graph where genes are connected to other genes based on similarity between expression pattern across samples (i.e. correlation score between gene A and B). Community detection will define modules as a group of genes that are densely connected to each other but sparsely connected to other genes in the network (within vs between edges). 

I can update this description to include this information

networks/2b_community_detection.py
ajlee21: Yes that is correct

networks/3_module_validation.py
ajlee21: I did play around with the parameters somewhat, but I can try some more and see what happens

networks/2b_community_detection.py
ajlee21: So the partition is either a `VertexClustering` or `VertexDendrogram` object that is essentially a dictionary mapping module ids or labels to vertex ids. This information can be combined with the graph which contains attributes for vertex and edges which then have vertex attributes like "name" and edge attributes like "weight". When we combine this information we can get the name of the vertex and which module id/label it maps to as well as the degree

PR #30, opened 2021-06-18
Calculate module composition

ajlee21:
This PR is calculating the composition of modules that were found by clustering (using affinity propagation) on the gene correlation matrix.

The main change in this PR:
1. `module_analysis/1_calc_module_composition.ipynb`, which inputs the gene modules and labels modules as mostly core, mostly accessory or mixed. I originally did this using a sampling approach that @jjc2718 and I discussed (i.e. comparing the number of accessory genes in a given modules vs the distribution of core genes that are sized matched to the number of accessory genes, since the number of accessory genes is much fewer compared to the number of core genes). The new code is now using a fisher's exact test, which I'm feeling a little shaky on my criteria to bin modules, so any feedback on that section would be really helpful
2. `2_add_annotations.ipynb` notebook that generates a summary dataframe with information about each gene including: the module membership, size of the module, the composition of the module, label of the module, and other annotations (i.e. gene name, KEGG pathways)

Note: I accidentally made a branch of a branch so the other files were already merged in from https://github.com/greenelab/core-accessory-interactome/pull/29

This is our first pass results calculating the composition. Moving forward, we will want to verify that these results are stable by re-running this analysis after we:
1. Update the input data to use the most recent QC'd version from our collaborators to try to correct for any spurious correlations
2. Generate multiple network partitions and aggregate the modules labels across partitions


module_analysis/1_calc_module_composition.py
```diff
@@ -0,0 +1,283 @@
...
+        # Do we want to include p-value criteria?
+        # Given the small sizes of the modules I think this is why
+        # most of the findings are not significant
+        # Use bonferrroni corrected p-value
+        if odds_ratio > 1 and pval < 0.05 / len(all_genes):
+            out_df.loc[module_id, "module label"] = "mostly core"
+        elif odds_ratio < 1 and pval < 0.05 / len(all_genes):
+            out_df.loc[module_id, "module label"] = "mostly accessory"
+        else:
+            out_df.loc[module_id, "module label"] = "mixed"
```
danich1, line 179: You could do the following (not here but something to keep in mind):
```python

data_rows = []
data_rows.append({
"module_id":module_id,
"num core":len(core_genes_in_module),
"num acc":len(acc_genes_in_module),
"odds ratio":odds_ratio,
"p-value":pval,
"module_label": (
"mostly core" if odds_ratio > 1 and pval < 0.05 / len(all_genes) else 
"mostly accessory" if odds_ratio < 1 and pval < 0.05 / len(all_genes) else
"mixed")
})

return pd.DataFrame.from_records(data_rows)
```
This builds the dataframe implicitly with each row as a dictionary and saves lines of code.

danich1: Another thing. It be helpful if you could include the total number of core and accessory genes outside your module. (aka fill out the full contingency table for each row.)

module_analysis/1_calc_module_composition.py
```diff
@@ -0,0 +1,283 @@
...
+plt.title("Size distribution of PAO1 mixed modules")
+
+f2 = sns.displot(
+    pao1_module_labels.loc[
+        pao1_module_labels["module label"] == "mostly accessory", "size"
+    ]
+)
+plt.title("Size distribution of PAO1 mostly accessory modules")
+
+f3 = sns.displot(
```
danich1, line 240: Is this suppose to be a blank plot? It showed up for me as blank and I want to confirm it isn't a bug.

module_analysis/1_calc_module_composition.py
```diff
@@ -0,0 +1,283 @@
...
+plt.title("Size distribution of PA14 mixed modules")
+
+g2 = sns.displot(
+    pa14_module_labels.loc[
+        pa14_module_labels["module label"] == "mostly accessory", "size"
+    ]
+)
+plt.title("Size distribution of PA14 mostly accessory modules")
+
+g3 = sns.displot(
```
danich1, line 259: Same comment as above.

module_analysis/2_add_annotations.py
```diff
@@ -0,0 +1,246 @@
...
+        gene in pao1_operons.loc[operon, "Genes"] for operon in pao1_operons.index
+    ]
+
+# Add operons to pathway annotations
+pao1_gene_annot = gene_to_pathways_df.merge(
+    gene_to_operons_df, left_index=True, right_index=True, how="outer"
+)
+
+print(pao1_gene_annot.shape)
+pao1_gene_annot.head()
```
danich1, line 170:
```suggestion
pao1_gene_annot.head().T
```
I recommend transposing the dataframe head so we can see the field names better.

danich1 (approved):
A few suggestions but overall looks good to me!

module_analysis/1_calc_module_composition.py
ajlee21: Nice! I do remember seeing this in your code, i like it!

module_analysis/1_calc_module_composition.py
ajlee21: Yes, thats correct. Since there are 0 modules labeled as mostly core, this plot is blank. I just have it plotting as I'll fiddle with different upstream methods and will just want to re-run this notebook

module_analysis/1_calc_module_composition.py
```diff
@@ -0,0 +1,283 @@
...
+                [len(core_genes_in_module), len(core_genes_outside_module)],
+                [len(acc_genes_in_module), len(acc_genes_outside_module)],
+            ]
+        )
+
+        # H0: The probability that the gene is core is the same
+        # whether or not you're in the module or outside
+        # H1: The probability that a gene is core is higher or lower inside the module
+        # than outside the module
+        # Should I do 2 one-sided tests, one for testing if
```
jjc2718, line 154: Did you try the one-sided tests? Curious if it makes any difference.

module_analysis/1_calc_module_composition.py
```diff
@@ -0,0 +1,283 @@
...
+        # Fill in df
+        out_df.loc[module_id, "num core"] = len(core_genes_in_module)
+        out_df.loc[module_id, "num acc"] = len(acc_genes_in_module)
+        out_df.loc[module_id, "odds ratio"] = odds_ratio
+        out_df.loc[module_id, "p-value"] = pval
+
+        # Do we want to include p-value criteria?
+        # Given the small sizes of the modules I think this is why
+        # most of the findings are not significant
+        # Use bonferrroni corrected p-value
```
jjc2718, line 173: If you're finding that this correction is too stringent, it would probably be fine to do a FDR correction here, since you're probably okay with having a few false positives. You can do this using [statsmodels](https://www.statsmodels.org/dev/generated/statsmodels.stats.multitest.multipletests.html) - I use the `fdr_bh` correction there in my work.

module_analysis/1_calc_module_composition.py
```diff
@@ -0,0 +1,283 @@
...
+# TO DO: Compare the composition across partitions
+# Save the matrix that maps module id to module label per partition
+# In another notebook look through module labels per partition
+# Calculate consistency of labeling using ARI for each pair of partitions
+# -
+
+# **Takeaway:**
+# * Most modules are mixed, some are mostly accessory. Only PA14 compendium have some mostly core modules
+# * PAO1 mixed and mostly accessory modules have similar sizes (~10 genes)
+# * PA14 mixed modules are ~10 genes, mostly accessory modules are smaller ~5 genes, mostly core modules are larger ~ 20 genes
```
jjc2718, line 283: You don't have any mostly core modules after multiple testing correction, right? Might be good to update this comment.

module_analysis/2_add_annotations.py
```diff
@@ -0,0 +1,246 @@
...
+
+print(pa14_gene_module_labels.shape)
+pa14_gene_module_labels.head()
+
+# ## Add KEGG pathways
+#
+# For each pathway, what genes are contained in it
+
+# +
+pao1_pathway_filename = "https://raw.githubusercontent.com/greenelab/adage/master/Node_interpretation/pseudomonas_KEGG_terms.txt"
```
jjc2718, line 115: Might be better to link to a specific commit here (instead of `master` since it could change).

jjc2718 (approved):
Looks good! The module binning and statistical testing implementation looks good to me (I made a few small comments but nothing major).

module_analysis/2_add_annotations.py
ajlee21: I see what you're saying. For now, I think I'll keep the columns as annotations because it'll be easier for our collaborators to sort when they are searching through these results

module_analysis/1_calc_module_composition.py
```diff
@@ -0,0 +1,283 @@
...
+        # -----|inside module|outside module
+        # core | # genes     | # genes
+        # acc  | # genes     | # genes
+        observed_contingency_table = np.array(
+            [
+                [len(core_genes_in_module), len(core_genes_outside_module)],
+                [len(acc_genes_in_module), len(acc_genes_outside_module)],
+            ]
+        )
+
```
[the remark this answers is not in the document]
ajlee21: So it turns out that the python implementation casts the table to an int so adding a pseudo count doesn't make a difference in this case: https://github.com/scipy/scipy/blob/44e31b1eb202ef91caad059952ba268d2610a60b/scipy/stats/stats.py#L4231

module_analysis/1_calc_module_composition.py
ajlee21: I did try running 1-sided and it didn't make a difference in the labeling

module_analysis/1_calc_module_composition.py
ajlee21: This is a good point. I'll add a params for this

module_analysis/1_calc_module_composition.py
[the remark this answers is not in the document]
danich1: Could you just add 1 to everything then and state that you added one to the counts for pvalue/odds_ratio calculations?

PR #31, opened 2021-07-05
common DEGs and core/acc annotations

ajlee21:
This PR uses SOPHIE-identified common DEGs and determines if those common DEGs are mostly core or accessory.

It appears that these common DEGs are mostly core, as expected.


common_genes/1_common_DEGs_and_gene_groups.ipynb
```diff
@@ -0,0 +1,892 @@
...
+   "id": "c51bf1f3",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Number of PAO1 core genes: 5366\n",
+      "Number of PA14 core genes: 5363\n",
+      "Number of PAO1 core genes in my dataset: 5361\n",
```
arielah, line 194: Why is there a difference in number of total core genes and number of core genes in your dataset? Could be worth a comment for clarity.

arielah (approved):
Looks good! Very neat, concise analysis so I didn't have much to recommend about it. Glad the results are what you expected :)

common_genes/1_common_DEGs_and_gene_groups.ipynb
ajlee21: Good point, I've added a comment to clarify this

PR #32, opened 2021-07-07
Re-run analysis using new QCd data

ajlee21:
This PR is re-running the analysis pipeline on a new cleaned version of the raw gene expression data. This new version of the data was carefully QC'd by our collaborator at Dartmouth to try to only include high quality samples and reduce the possibility of spurious correlations that might be due to noise in the data.

There are many files changed since I was re-running the analysis and generating new result files, but these do not need a heavy review if at all. The main new change to be reviewed is the notebook `module_analysis/3_calc_prob.ipynb` to calculate the conditional probability that given a gene is accessory, what is the likelihood that gene is found in an accessory module.


module_analysis/3_calc_prob.py
```diff
@@ -0,0 +1,170 @@
...
+# Currently there are no mostly core modules, so this is a placeholder
+# Get number of core genes in "mostly core" modules
+"""num_core_gene_in_core_mod_pao1 = pao1_subset.loc[
+    pao1_subset["module label"] == "mostly core", "num core in module"
+].sum()
+num_core_gene_in_core_mod_pa14 = pa14_subset.loc[
+    pa14_subset["module label"] == "mostly core", "num core in module"
+].sum()"""
+
+lik_pao1_acc = num_acc_gene_in_acc_mod_pao1 / (len(pao1_core) + len(pao1_acc))
```
jjc2718, line 108: I think you could just do this in one step if you want, since the total numbers of genes will cancel out. Here's the logic for how I got there (see if you agree with it), based on the formula at the top of the notebook:

Image from iOS (5)

This should give the same answer as what you have, but might save you a few lines of code - unless you want to keep looking at the prior distribution as part of the notebook (which I'm totally fine with).

module_analysis/3_calc_prob.py
```diff
@@ -0,0 +1,170 @@
...
+print(pr_pa14_core)
+
+# ## Calculate conditional probability
+
+# +
+pr_acc2acc_pao1 = lik_pao1_acc / pr_pao1_acc
+pr_acc2acc_pa14 = lik_pa14_acc / pr_pa14_acc
+
+print(
+    f"Probability of accessory gene being co-expressed with another accessory gene in PAO1 is {pr_acc2acc_pao1}"
```
jjc2718, line 154: So what's your interpretation of this? Does this mean that in PAO1 a little less than half of accessory genes are co-expressed primarily with other accessory genes, and the other half of accessory genes tend to be in non-specific modules (i.e. with a mixture of both core and accessory genes)?

Are you surprised at these numbers, or are they close to what you expected to see?

module_analysis/3_calc_prob.py
```diff
@@ -0,0 +1,170 @@
...
+pao1_core = core_acc_dict["core_pao1"]
+pa14_core = core_acc_dict["core_pa14"]
+pao1_acc = core_acc_dict["acc_pao1"]
+pa14_acc = core_acc_dict["acc_pa14"]
+
+# ## Calculate likelihood
+# $Pr(\text{gene x in acc module}\cap \text{gene x is acc gene})$ is the number of accessory genes in accessory modules
+
+# Get subset of df
+pao1_subset = pao1_membership[["module id", "module label", "num acc in module"]]
```
jjc2718, line 83: Are these modules just from one run of community detection? Or are these an aggregate over multiple community detection runs (different seeds, different methods, etc)?

jjc2718 (approved):
Nice work! Just a few small comments, mostly for my own clarification.

module_analysis/1_calc_module_composition.py
```diff
@@ -154,13 +168,15 @@ def label_modules(
         )
 
         # Add 1 to avoid inf
-        observed_contingency_table = observed_contingency_table + 1
+        observed_contingency_table = observed_contingency_table
```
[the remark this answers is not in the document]
ajlee21: The pseudocount just got rid of the inf odds ratios, but doesn't change the criteria to label modules, so I think I can remove it

processing/1_create_compendia.py
```diff
@@ -131,22 +134,22 @@
...
+        "median_acc_expression_pao1<@opp_threshold & median_acc_expression_pa14>@same_threshold"
     ).index
 )
 
 # +
 # Check that there are no samples that are binned as both PAO1 and PA14
 shared_pao1_pa14_binned_ids = list(set(pao1_binned_ids).intersection(pa14_binned_ids))
 
-assert len(shared_pao1_pa14_binned_ids) == 0
+# assert len(shared_pao1_pa14_binned_ids) == 0
```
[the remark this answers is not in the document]
ajlee21: I commented this out when I was testing different thresholds, but this assertion should be kept

module_analysis/3_calc_prob.py
ajlee21: NICE MATH! I'm ok keeping it this way so its more intuitive for future me:)

module_analysis/3_calc_prob.py
ajlee21: So to me this means that accessory genes are no more likely to be co-expressed with other accessory genes as with other core genes. However, given the small size of our modules and the high bias in core genes, we are trying to explore other approaches to calculate this probability (stay tuned)

module_analysis/3_calc_prob.py
ajlee21: This is a single community detection run. Once we settle on an approach, I will want to repeat this for multiple seeds.

module_analysis/1_calc_module_composition.py
```diff
@@ -207,8 +223,8 @@ def label_modules(
         )
 
         # Add columns
+        out_df["FDR corrected p-value"] = cpva
```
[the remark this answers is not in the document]
ajlee21: Good catch, thats a typo

PR #34, opened 2021-07-29
Refactor to prepare for separate module analysis

ajlee21:
This PR refactors the repo to prepare for the new set of analyses we will perform where we'll separate between core and accessory genes and examine expression patterns in these two different groups.

Most changes are renaming and archiving old files. Some code changes include

1. Add a venn diagram to visualize the comparison between core genes and common DEGs.
2. Add ability to select core or accessory gene subsets to cluster on


location_vs_expression_relationships/1_genome_dist_vs_expression_dist.py
```diff
@@ -59,23 +59,24 @@
 )
 # -
 
-# ### Import module memberships
+# ### Import gene ids
```
danich1, line 62: Are the two functions in this notebook doing similar things? If so could it be collapsed into one function?

danich1 (approved):
LGTM! Just one quick question as I'm thinking about code refactor.

location_vs_expression_relationships/1_genome_dist_vs_expression_dist.py
ajlee21: So they're both finding relationships but in different spaces - one is genome space (so who is your nearest neighbor, core or acc?) and the other is in expression space (who are you most co-expressed with, core or acc?). I think conceptually they are similar but the way I need to calculate things is a little different. Since the functions are already fairly large I'd rather keep them separate, mainly for my own understanding I think having them separate helps. But good question!

PR #35, opened 2021-08-05
Add analysis to compare core-core modules across strains

ajlee21:
This notebook examines how stable core genes are across strains PAO1, PA14. Here we start with a given PAO1 gene and rank who its related to from most to least correlated. We then ask if the homologous PA14 gene as the same ranking. If they do, then this gene is considered stable.

The approach used it to:
1. Make core-core correlation matrix
2. For given core gene A, get rank of how other genes are correlated to A in PAO1. Do the same in PA14.
3. Are the correlations for homologous genes correlated?
4. Each gene will have a correlation score
5. Which genes have the most similar transcriptional relationships (i.e., highest correlation)? Which are the least? What does the distribution of stabilities look like?

Looks like the most stable core genes are related to the secretion system that plays a role in _P. aeruginosa_ virulence and resistance, while the less stable core genes are related to metabolism and other chemical reactions. I will need to think more about potential next analyses and what question to ask about core-core modules.


nrosed:
looks good! The notebook is clear and the figures nicely answer the questions you lay out at the beginning of the notebook.
Very small questions/comments:
- how is correlation calculated in your initial correlation matrix? I'm guessing person because I see that you use it in `compare_gene_relationships`?
- If I understand correctly, all pa14/pao1 homologs are significantly correlated? I'm just thinking if it makes sense to restrict the low_pa14/low_pao1 to p-values that are not significant. I guess it will depend on what you are interested in, but probably filtering by correlation is better than p-value.
    - Follow-up question, are core gene homologs more correlated than non-homologs? Just wondering what the expected rate of correlation is.

nrosed (approved):

ajlee21:
> looks good! The notebook is clear and the figures nicely answer the questions you lay out at the beginning of the notebook.
> Very small questions/comments:
> 
> * how is correlation calculated in your initial correlation matrix? I'm guessing pearson because I see that you use it in `compare_gene_relationships`?

The correlation matrix was generated using the notebooks in [this directory](https://github.com/greenelab/core-accessory-interactome/tree/993682730aca383ba628a7a2a7204375ef755c21/correlation_modules). We log-transformed and applied SPELL to the raw expression dataset. Then we took the Pearson correlation of that transformed matrix. The rationale for the processing can be found in the readme file of that directory, but basically it was done to balance out the signals that are in the dataset.

> * If I understand correctly, all pa14/pao1 homologs are significantly correlated? I'm just thinking if it makes sense to restrict the low_pa14/low_pao1 to p-values that are not significant. I guess it will depend on what you are interested in, but probably filtering by correlation is better than p-value.

So all the core genes (i.e. those with pao1/pa14 homologs) have similar transcriptional profiles across strains. So there are some core genes that are highly consistent across strains (i.e. they have a similar correlation profile/are highly related to the same set of core genes in PAO1 and PA14), these are the ones on the right tails. I'll try to clarify my comment on this more.

>   * Follow-up question, are core gene homologs more correlated than non-homologs? Just wondering what the expected rate of correlation is.

So since core genes have homologs we were able to do this analysis. Those genes we cannot compare how consistent non-homlogous/accessory genes are. If I understand your question correctly

PR #36, opened 2021-08-05
Get acc-acc modules and annotate

ajlee21:
This PR adds a notebook to annotate accessory-accessory modules with information that can be used to narrow down which modules to explore


acc_acc_modules/1_annotate_acc_modules.py
```diff
@@ -0,0 +1,373 @@
...
+import pandas as pd
+import numpy as np
+from itertools import product
+from core_acc_modules import paths
+
+random.seed(1)
+# -
+
+# User param
+method = "affinity"
```
arielah, line 45: You say "user param", will a future user want to change this? If so, what are the supported options?

acc_acc_modules/1_annotate_acc_modules.py
```diff
@@ -0,0 +1,373 @@
...
+pa14_gene_annot_filename = paths.GENE_PA14_ANNOT
+
+pao1_gene_annot = pd.read_csv(pao1_gene_annot_filename, index_col=0, header=0)
+pa14_gene_annot = pd.read_csv(pa14_gene_annot_filename, index_col=0, header=0)
+# -
+
+# Import metadata of samples
+metadata_filename = paths.SAMPLE_METADATA
+
+pao1_gene_annot = pao1_gene_annot["Name"].to_frame("gene name")
```
arielah, line 71: This is really nitpicky, but what is this code block doing? It's not clear from the functions themselves.

acc_acc_modules/1_annotate_acc_modules.py
```diff
@@ -0,0 +1,373 @@
...
+# Add gene names
+pao1_gene_module_labels = pao1_membership.merge(
+    pao1_gene_annot, left_index=True, right_index=True
+)
+pa14_gene_module_labels = pa14_membership.merge(
+    pa14_gene_annot, left_index=True, right_index=True
+)
+
+print(pao1_gene_module_labels.shape)
+pao1_gene_module_labels.head()
```
arielah, line 85: Are you expecting this many NaNs? I would put in a comment about that, because if I was running this notebook as a future user, that would look like a problem to me.

acc_acc_modules/1_annotate_acc_modules.py
```diff
@@ -0,0 +1,373 @@
...
+print(pa14_gene_module_labels.shape)
+pa14_gene_module_labels.head()
+
+# ## Add expression information
+#
+# 1. What is the baseline level of expression for each gene in the module?
+# 2. What is the expression level of genes in a clinical context (i.e. clinical samples)?
+
+# Read in expression data
+pao1_compendium = pd.read_csv(paths.PAO1_COMPENDIUM, sep="\t", index_col=0)
```
arielah, line 96: Pretty sure columns are gene names, rows are module names? Where does this ERX5* naming come from?

acc_acc_modules/1_annotate_acc_modules.py
```diff
@@ -0,0 +1,373 @@
...
+
+# Calculate median expression across all samples
+pao1_median_all = pao1_compendium.median().to_frame("median expression")
+pa14_median_all = pa14_compendium.median().to_frame("median expression")
+
+pao1_median_all.head()
+
+
+# Select subset of samples and calculate the median expression across that subset of samples
+# TO DO: Move this into utils
```
arielah, line 109: Do you want to do this now?

acc_acc_modules/1_annotate_acc_modules.py
```diff
@@ -0,0 +1,373 @@
...
+# +
+# Subset compendium
+# subset_pao1_compendium = pao1_compendium.loc[selected_sample_ids]
+# subset_pa14_compendium = pa14_compendium.loc[selected_sample_ids]
+
+# +
+# print(subset_pao1_compendium.shape)
+# print(subset_pa14_compendium.shape)
+
+# +
```
arielah, line 156: Just want to confirm that the notebook will run without errors with these sections commented out. I'm sure it's fine, I've just run into problems with that before.

acc_acc_modules/1_annotate_acc_modules.py
```diff
@@ -0,0 +1,373 @@
...
+        # print("module", grp_name)
+
+        # Trim off "PA" and convert number to integer
+        ids = grp_df.index
+
+        # Convert trailing id numbers to floats
+        num_ids = [float(_id.split(pa_prefix)[1]) for _id in ids]
+
+        abs_dist = []
+        for gene1, gene2 in product(num_ids, num_ids):
```
arielah, line 208: This is probably just ignorance on my part, but what does `product()` do here? I'm not sure how you're getting genomic locations out of this.

acc_acc_modules/1_annotate_acc_modules.py
```diff
@@ -0,0 +1,373 @@
...
+pao1_gene_annot = pao1_gene_annot.merge(
+    pao1_module_dist, left_index=True, right_index=True, how="left"
+)
+pa14_gene_annot = pa14_gene_annot.merge(
+    pa14_module_dist, left_index=True, right_index=True, how="left"
+)
+
+# ## Add KEGG pathways
+#
+# For each pathway, what genes are contained in it. This information is only available for PAO1.
```
arielah, line 253: Does this mean you have info for the core genes and the PA01 accessory genes? Wouldn't the core gene modules be the same across PA01 and PA14?

acc_acc_modules/1_annotate_acc_modules.py
```diff
@@ -0,0 +1,373 @@
...
+# This operon df contains annotations from predicted operons based on DOOR database
+# predictions which make up the majority of the operons) as well as some that
+# are curated (i.e. PseudoCAP)
+# There are some that have multiple PseudoCAP annotations too
+
+# Here we will keep the last PseudoCAP annotations
+# Note: Do we want to discard these annotations all together
+# or will these need to be carefully curated to determine which to keep?
+# We will use the curated annotation here
+pao1_operon = pao1_operon[~pao1_operon.index.duplicated(keep="last")]
```
arielah, line 310: Are you sure the curated annotation will always be the last here?

acc_acc_modules/1_annotate_acc_modules.py
arielah: Just realized this whole thing is only focusing on accessory genes. Nvm!

arielah (approved):
LGTM! My comments are mostly about idiotproofing for future users like me haha. Interesting work!

acc_acc_modules/1_annotate_acc_modules.py
ajlee21: Very good point I will add text here

acc_acc_modules/1_annotate_acc_modules.py
ajlee21: `pao1_gene_annot["Name"]`: is selecting column "Name" from dataframe `pao1_annot`, which returns a pandas series, which I then convert to a dataframe by calling `to_frame()` which the column name "gene name". This will just allow me to do my merges that are in the notebook

acc_acc_modules/1_annotate_acc_modules.py
ajlee21: Yes this code allows me to select a subset of samples of interest to get expression from, but I couldn't find a subset that I had expression data for yet, so they are commented out for now in case we are able to find a subset after talking with my collaborators. I will make a note of this

acc_acc_modules/1_annotate_acc_modules.py
ajlee21: So this dataframe is sample id x gene id, the sample id's are from SRA. I'll add this as a note too

acc_acc_modules/1_annotate_acc_modules.py
ajlee21: Nice catch! I forgot I wanted to do this. I'll update this so the notebook is cleaner

acc_acc_modules/1_annotate_acc_modules.py
ajlee21: Good point. when i looked at the DB it looked like it was sorted, but I can also just sort before to make sure that this is the case

acc_acc_modules/1_annotate_acc_modules.py
ajlee21: So `num_ids` is a list of the the gene ids, which we're using as our genomic location. Say our list is ["PA0001", "PA0004", "PA0010"], then product will return an iterator where at the first iteration gene1 = "PA0001" and gene2="PA0001" so there is a distance of 0 here. But in the next iteration we have gene1 = "PA0001" and gene2 = "PA0004"

PR #37, opened 2021-08-25
Validate that correlation modules are biologically relevant

ajlee21:
This PR updates the validation of the correlation modules generated in a few ways:

1. Examines the coverage of co-operonic/regulonic genes across modules. We would expect that the probability that a pair of genes is in the same module given they are in the same regulon to be higher using the true module labels compared to the shuffled labels, which we see below.
image

2. We compared the composition of modules across the array and RNA-seq compendia. We would expect consistency, which we do see based on the distribution of p-values that indicate how well modules from the array compendium mapped to the RNA-seq compendium. Here the distribution of p-values is lower using the true module labels compared to the shuffled labels.
image

3. We looked for enrichment of modules in KEGG pathways. We would expect some modules to correlate with KEGG pathways. The result we got isn't what I would expect, maybe this isn't the best way to assess this though.
 
image


correlation_modules/3_module_validation.py
```diff
@@ -317,258 +292,293 @@ def plot_dist_modules(clustering_method_list):
...
+)
+fig_operon = sns.histplot(
+    pa14_operon_shuffle_prob.loc[
+        pa14_operon_shuffle_prob["pr(x,y in module|x,y in operon)"] > 0,
+        "pr(x,y in module|x,y in operon)",
+    ],
+    bins=bins_shared,
+    color="grey",
+    ax=axes[1],
+    label="shuffle",
```
ben-heil, line 519: Isn't it weird that these probabilities are so low? Wouldn't you expect genes in the same operon to be really tightly coregulated?

correlation_modules/3a_validate_module_composition.py
```diff
@@ -0,0 +1,463 @@
...
+    y=rnaseq_to_array_combined["p-value"],
+    palette="Blues",
+    notch=True,
+)
+g.set_title("RNA-seq to array modules")
+
+# **Takeaway:**
+#
+# * Comparing how RNA-seq modules map to array modules, looks like most modules have very low p-values, meaning they map well between RNA-seq and array.
+# * In contrast, random modules have slightly higher p-values, indicating that the modules don't map as well, though the p-values are still very low.
```
ben-heil, line 360: Why do the random modules have low p-values? Is it because genes' expression is highly correlated?

ben-heil (approved):
Cool results!

1) To what extent does clustering reflect the biology of PA gene expression? Does the assumption that each gene should live in one module hold? (This isn't a criticism, I legitimately have no idea what the structure of bacterial gene expression regulation looks like/how strong regulons are as compared to operons).

2) If you're looking for another clustering method at some point, spectral clustering might be a good fit. It thinks about clustering as cutting up a similarity network, which at least aesthetically is similar to trying to partition a gene coexpression network

ajlee21:
> Cool results! Sorry for the late review, I totally missed this notification yesterday.
> 
> A few comments:
>
> 1. To what extent does clustering reflect the biology of PA gene expression? Does the assumption that each gene should live in one module hold? (This isn't a criticism, I legitimately have no idea what the structure of bacterial gene expression regulation looks like/how strong regulons are as compared to operons).

Very good question! There is this other method that weights a gene's contribution to different modules so that genes are not exclusively belonginging to a single cluster, which I think is a simplified assumption that people make for these types of analyses. https://arxiv.org/abs/2106.00657

We were considering using this method for our work, but it requires integer values as input right now, but looks like they are planning to extend this soon.

> 2. If you're looking for another clustering method at some point, spectral clustering might be a good fit. It thinks about clustering as cutting up a similarity network, which at least aesthetically is similar to trying to partition a gene coexpression network

Thanks! I'll make a note of that

correlation_modules/3_module_validation.py
ajlee21: Looks like I didn't select the best matching module and operon pair, I was accidentally plotting all p-values for all mappings. I just fixed the code and there looks like there is more of a shift towards higher probabilities, though there are still some operons that have a fairly low probability of being in the same module. Based on the `describe` statistics, I don't see much that stands out. But I think its at least a good sign that most operons are clustered in the same module

correlation_modules/3a_validate_module_composition.py
ajlee21: Very good catch! After correcting for multiple tests, I get trends that are more expected (i.e. low corrected p-values for true labels and high corrected p-values for shuffled labels). However, if I examine the overlap between modules, all of the best mapping based on corrected p-value scores, are 0. This behavior is unexpected to me - I wonder if this test is in appropriate given the rarety of the event. From my reading, I think I've implemented this test correctly. I made a note in this notebook about this and I'll have to return to this. This analysis of validating that our modules are capturing relevant biology was more critical when we were planning to do analyses on these modules but with our new story, it looks like this step isn't too necessary.

PR #39, opened 2021-09-01
Minor refactor and quantify common DEG analysis

ajlee21:
This PR has a couple of small changes:

1. Some refactoring: renaming files, pulling out functions from notebooks into scripts
2.  Calculate enrichment of common DEGs in core and accessory genes: `1_common_DEGs_and_gene_groups.ipynb`


danich1 (approved):
LGTM! If I'm understanding correctly, your chi square test is saying that genes have an equal chance being of being in core or not in core regardless of the module?  Is this the same for all modules (assuming there is more than one)?

ajlee21:
> If I'm understanding correctly, your chi square test is saying that genes have an equal chance being of being in core or not in core regardless of the module? Is this the same for all modules (assuming there is more than one)?

So here I was testing to see if there was a significant association between common DEGs vs core/accessory genes. Using the Fisher's exact test, it looks like there was an odds ratio greater than 1, indicating that there is a positive relationship between common DEGs and a gene being core, however the p-value was not significant in this case due to a skewing in the distribution of categories. 

This analysis was applied to all genes, not a single module. I'll try to clarify my comments in the notebook.

PR #40, opened 2021-09-03
Add enrichment analysis of modules

ajlee21:
This PR is updating a few different analyses:

1.  Add KEGG pathway enrichment to find acc-acc modules that have an over-representation of some KEGG pathway. (`acc_acc_analysis/1_annotate_acc_modules.ipynb`)
2.  Look for which KEGG pathways the most and least stable core genes are found to be associated (`core_core_analysis/2_find_KEGG_associations.ipynb`)
3.  Find which genes the most and least stable core genes are related to (`core_acc_analysis/stable_gene_relationships.ipynb`)


core_acc_analysis/stable_gene_relationships.py
```diff
@@ -0,0 +1,397 @@
...
+    y="total",
+    hue="gene type",
+    ax=axes[1],
+    palette=sns.color_palette("Paired"),
+)
+fig.legend_.remove()
+fig.set_title("Starting with least stable core gene PAO1")
+fig.set_ylabel("Number of genes")
+fig.set_xlabel("Rank correlation in expression space")
+
```
nrosed, line 329: small question -- are you interested in the proportion of genes that are highly correlated, or the absolute number? I'm just wondering is a stacked proportional bar chart would be better here because then you might be able to plot more modules, but not sure how important that is.

nrosed:
Looks good! no major comments from me, everything seems clear and straightforward. 👍🏼

nrosed (approved):

core_acc_analysis/stable_gene_relationships.py
ajlee21: This is a great point! My next step is to figure out what is the best way to visualize this data because, I agree, I think what would be better is if there is some sort of normalization of the data so that we can see proportionally who is related to who.

PR #41, opened 2021-09-10
Create figures and data for Georgia's manuscript

ajlee21:
This PR is creating figures and data that will be used in Georgia's manuscript, which describes the Pseudomonas aseruginosa RNA-seq compendia we generated. This manuscript will allow us to focus our manuscript on the core-accessory analysis work we've been doing.

This PR contains 2 major changes:
1. Updated the colors and font of the gene expression plots in `explore_data/cluster_by_accessory_gene.ipynb` to use Georgia's theme
2. Create a dataframe that annotates pre-binned samples with their median accessory expression and strain labels from SRA and our approach: `processing/1_create_compendia.ipynb`

Most files changed are figures.


processing/1_create_compendia.py
```diff
@@ -191,6 +191,61 @@
...
+pao1_pa14_acc_expression_label.loc[pa14_binned_ids, "Our label"] = "PA14-like"
+
+pao1_pa14_acc_expression_label.head()
+
+# +
+# Confirm dimensions
+pao1_expression_prebin_filename = paths.PAO1_PREBIN_COMPENDIUM
+pa14_expression_prebin_filename = paths.PA14_PREBIN_COMPENDIUM
+
+pao1_expression_prebin = pd.read_csv(
```
arielah, line 230: Is the point of splitting up into the two expression prebins just to confirm they're the same size? I think so because you're only saving the combined one. Could you just check sample sizes without having to load in the other files?

arielah (approved):
Looks good! Pretty straightforward. Exciting that this paper is close!

processing/1_create_compendia.py
ajlee21: So the two expression prebins are because the same samples were mapped to 2 different references - a PAO1 and a PA14 reference. So this assertion is to make sure that the number of samples is the same in both, which it is, and that when we added information about our accessory gene expression and labels we retained the same number of samples, which we did. I'll add a comment about this to clarify

PR #46, opened 2021-10-08
Add scripts and analysis nbs from SOPHIE

ajlee21:
This PR is adding the analysis to apply SOPHIE to find common DEGs within the PAO1 and PA14 compendia. This code was already reviewed in https://github.com/greenelab/generic-expression-patterns/pull/96 but here we're moving it into this repository as opposed to being in the SOPHIE repo.


ajlee21:
> Looks good to me!
> 
> So if you update the code in `generic_expression_patterns`, will you have to update the code here too? Or are you okay with just using this version of the generic genes code for all of the experiments in this repo?

At this point, I don't anticipate any major changes to the code in the `generic_expression_patterns` repo so this should be ok.

PR #48, opened 2021-10-29
Compare SPELL vs counts correlation

ajlee21:
This PR adds an experiment to determine which correlation matrix to use: counts vs SPELL-processed counts version.

This PR contains the following major changes:
1. Add `spell_vs_counts_experiment/README.md` describes the experiment
2. Add `spell_vs_counts_experiment/1a_compare_SPELL_vs_counts_correlation.ipynb` performs the experiment
3. Pull out previous analysis comparing most stable core genes using SPELL vs counts into its own notebook `3_core_core_analysis/compare_most_stable_genes.ipynb`
4. This PR also found a big in my code. Previously we were generating a correlation matrix using all the genes, but now we were making two different correlations - one using only core genes, another using only accessory genes. When I added this  new feature I made it so that the correlation was created first and then I used pandas to split the correlation matrix into the different gene subsets. However this meant that the correlation scores in both the core-core and acc-acc matrices are based on **all genes** not just core or accessory. This means that any relationships between core-core or acc-acc using the correlation matrix were based on SVs that represent linear combinations of all genes. Since we're examining the gene groups separately and then together I think it makes sense to split the genes _before_ the correlation is calculated. This change is added to the `1_correlation_analysis.ipynb`

The remaining files changed are data files created and others that were re-organized


2_correlation_analysis/spell_vs_counts_experiment/README.md
```diff
@@ -0,0 +1,23 @@
+# Compare SPELL vs counts correlation
+
+There are two possible correlation matrices we can use for our analysis:
+
+1. Correlation of the MR counts expression matrix
+2. Correlation of the MR counts that have been processed using SPELL
```
jjc2718, line 6:
```suggestion
1. Correlation of the median-ratio counts expression matrix
2. Correlation of the median-ratio counts that have been processed using SPELL
```
I'd suggest spelling this out here, in case people haven't read your other scripts and documentation.

jjc2718, line 6: Also, could you add a link to a paper or something describing SPELL here?

2_correlation_analysis/spell_vs_counts_experiment/README.md
```diff
@@ -0,0 +1,23 @@
...
+The correlation of the SPELL matrix relates genes based on the gene coefficient matrix - **relate genes over their contribution to singular vectors (linear combination of genes - linear relationship between genes)**.
+High correlation means that a pair of genes contributes similarly to a singular vector, which are the axes pointing in the direction of the spread of the data and capture how genes are related to each other
+* Pro: Gene contributions are more balanced so that redundant signals (i.e. many genes from the same pathway - genes that vary together) are represented by a few SVs as opposed to many samples. More balanced also means that more subtle signals can be amplified (i.e. genes related by a smaller pathway are also captured by a few SVs)
+* Con: Can amplify noise - i.e. an SV that corresponds to some technical source of variability now has a similar weight to other real signals
+
+For more information comparing using counts vs SPELL-processing see: https://docs.google.com/presentation/d/18E0boNODJaxP-YYNIlccrh0kASbc7bapQBMovOX62jw/edit#slide=id.gf9d09c6be6_0_0
```
jjc2718, line 21: These Google slides seem to have restricted access (I can't see them).

2_correlation_analysis/spell_vs_counts_experiment/README.md
```diff
@@ -0,0 +1,23 @@
...
+* Pro: Easy to interpret
+* Con: Many gene pairs found to have a high correlation because many genes are related to the same pathway have the similar expression profiles. This is consistent with [Myers et al.](https://link.springer.com/article/10.1186/1471-2164-7-187), who found that there can be an over-representation of genes associated with the same pathway (i.e. a large fraction of gene pairs represent ribosomal relationships). This very prominent signal makes it difficult to detect other signals. Figure 1C demonstrates that a large fraction of gene pairs are ribosomal relationships - in the top 0.1% most co-expressed genes, 99% belong to the ribosome pathway. Furthermore, protein function prediction based on co-expression drop dramatically after removing the ribisome pathway (Figure 1A, B).
+
+To try to remove this very dominant global signal in the data. Here we are applying dimensionality reduction techniques in addition to scaling the data using a method called SPELL.
```
jjc2718, line 15: This was sort of the same idea behind using ADAGE to construct networks in the generic-expression project, right? To remove the strong operon signal that tends to dominate co-expression networks?

I guess maybe you could do something similar here, but it seems like SPELL is getting at the same idea so it's probably fine.

2_correlation_analysis/spell_vs_counts_experiment/1_correlation_analysis.py
```diff
@@ -0,0 +1,253 @@
...
+# +
+# Params
+
+# Which subset of genes to consider: core, acc, all
+subset_genes = "all"
+
+# The number of accessory genes is 200 - 500
+# The number of core genes is ~ 5000
+# So the number of singular vectors is relative to the number of genes
+if subset_genes == "acc":
```
jjc2718, line 51: How did you choose 50 and 100 here? Could you look at the variance explained by this many SVs and make sure they're similar for each dataset?

2_correlation_analysis/spell_vs_counts_experiment/1_correlation_analysis.py
```diff
@@ -0,0 +1,253 @@
...
+elif subset_genes == "acc":
+    pao1_compendium = pao1_compendium[pao1_acc]
+    pa14_compendium = pa14_compendium[pa14_acc]
+
+print(pao1_compendium.shape)
+print(pa14_compendium.shape)
+
+# ## Correlation of raw gene expression data
+#
+# Here is the correlation of the raw data without any malnipulations. This will serve as a reference to compare the correlations below where applied corrections to the correlations to account for the dominant signal described above.
```
jjc2718, line 113:
```suggestion
# Here is the correlation of the raw data without any manipulations. This will serve as a reference to compare the correlations below where applied corrections to the correlations to account for the dominant signal described above.
```

2_correlation_analysis/spell_vs_counts_experiment/1_correlation_analysis.py
```diff
@@ -0,0 +1,253 @@
...
+#
+# From [Hibbs et. al.](https://academic.oup.com/bioinformatics/article/23/20/2692/229926), we apply their "signal balancing technique that enhances biological information". This is the first part of their [SPELL](https://spell.yeastgenome.org/) algorithm that is described in section 2.3.1. SPELL calculates the correlation on the gene coefficient matrix, $U$ (i.e. how much genes contribute to a latent variable) that is generated after applying SVD. This matrix represents how genes contribute to independent latent variables that capture the signal in the data where the variance of the variables is 1. The idea is that correlations between gene contributions are more balanced so that less prominent patterns are amplified and more dominant patterns are dampended due to this compression. Figure 3 shows how well SPELL recapitulates biology (i.e. the relationship between genes within a GO term) compared to Pearson correlation.
```
jjc2718, line 167: Oh, now I see how this works. It does sound very similar to using ADAGE to calculate correlation over all the latent variables (except in this case you're using linear "latent variables" or singular vectors). This is a nice description!

2_correlation_analysis/spell_vs_counts_experiment/1_correlation_analysis.py
```diff
@@ -0,0 +1,253 @@
...
+)
+
+# Save
+pao1_log_spell_filename = os.path.join(
+    paths.LOCAL_DATA_DIR, f"pao1_{subset_genes}_log_spell_clustermap_test.png"
+)
+h1a.savefig(pao1_log_spell_filename, dpi=300)
+
+# +
+h2a = sns.clustermap(pa14_corr_log_spell, cmap="BrBG", center=0, figsize=(20, 20))
```
jjc2718, line 221: These do look very different than the heatmaps without the correction. I suspect you'd find many more communities if you did community detection on this matrix, and the sizes would be much more uniform (not sure if this is in your plans or not, just something I'd be curious about).

2_correlation_analysis/spell_vs_counts_experiment/1a_compare_SPELL_vs_counts_correlation.py
```diff
@@ -0,0 +1,441 @@
...
+    notch=True,
+    ax=axes3[1],
+)
+axes3[0].set_title(f"counts matrix (threshold={pao1_corr_counts_threshold: .2f})")
+axes3[1].set_title(f"SPELL matrix (threshold={pao1_corr_spell_threshold: .2f})")
+axes3[0].set_ylabel("% edges")
+axes3[1].set_ylabel("% edges")
+axes3[0].set_xlabel("")
+axes3[1].set_xlabel("")
+plt.suptitle(
```
jjc2718, line 439: This is a nice experiment! I'm convinced that using SPELL makes sense 👍

3_core_core_analysis/1_core_core_relationships_across_strains.py
```diff
@@ -244,11 +249,10 @@ def compare_gene_relationships(gene_mapping_dict, mapping_to, pao1_corr, pa14_co
 
 # ## Plots
 
-# +
-# Add label for most and least stable genes
+# Add label for most and least stable genes using threshold cutoff
 pao1_corr_df["label"] = ""
 pa14_corr_df["label"] = ""
-pao1_corr_df.loc[
+"""pao1_corr_df.loc[
```
jjc2718, line 255: Do you want to keep this commented-out code?

jjc2718 (approved):
Looks good! A few comments.

2_correlation_analysis/spell_vs_counts_experiment/README.md
ajlee21: Yes that is the same idea here you're right. ADAGE would be another way to have approached this. I think SPELL was a simple first step especially for our downstream interpretations

2_correlation_analysis/spell_vs_counts_experiment/1_correlation_analysis.py
ajlee21: This was based on a manual inspection of the heatmaps -- examining the size of the clusters and make sure the very dominant signal was removed. I can also look at the variance explained here and confirm that the results are similar to the numbers chosen

2_correlation_analysis/spell_vs_counts_experiment/1_correlation_analysis.py
ajlee21: We're planning to perform a clustering analysis on the accessory-accessory correlation matrix to extract accessory modules. But for the core-core correlation matrix, we will be performing an analysis using the correlation matrix directly. Stay tuned!

2_correlation_analysis/spell_vs_counts_experiment/1_correlation_analysis.py
jjc2718: Up to you, if you've already looked at cluster size and confirmed the results are what you expect then maybe there's no real need.

3_core_core_analysis/1_core_core_relationships_across_strains.py
ajlee21: No this isn't needed anymore

2_correlation_analysis/spell_vs_counts_experiment/1_correlation_analysis.py
ajlee21: I've added a comment to clarify how these were chosen. Its pretty quick to plot that so I think I'll just add it just to make sure that the numbers align with my observations

PR #49, opened 2021-11-04
Refactor and redo analyses using SPELL correlation matrix

ajlee21:
This PR is a follow-up from https://github.com/greenelab/core-accessory-interactome/pull/48, where we compared using two different correlation matrices. From that PR we decided to use the SPELL-processed correlation matrix and here we are re-running our analyses and cleaning up the directories.

Most of the files changed were previously reviewed and just re-run.

Main code changes are plots that I updated to be more clear including:
1. Update heatmap in `2_correlation_analysis/1_correlation_analysis.ipynb` to include positive and negative values
2. Clean up displot of correlation scores `3_core_core_analysis/1_core_core_relationships_across_strains.ipynb`. Takeaway: there exists some core genes that are very stable across strains and others that are not.
3. Update barplot in ` 5_core_acc_analysis/stable_gene_relationships.ipynb` to use normalized version of the counts to visualize trends better. Takeaway: least stable core genes are more co-expressed with accessory genes than expected. Note: I also started to make updates to `scripts/gene_relationships.py` to add a confidence interval to this plot but I will need to finish this in the next PR.

If you have any thoughts on how to improve any of these figures to reflect the takeaways, that would be helpful


3_core_core_analysis/5_KEGG_enrichment_of_stable_genes.py
```diff
@@ -144,17 +146,86 @@ def KEGG_enrichment_of_stable_genes(similarity_score_df, gene_list, kegg_df):
...
+plt.legend(
+    title="corrected p-value",
+    title_fontsize=14,
+    fontsize=12,
+    bbox_to_anchor=(1.05, 0.8),
+    loc="upper left",
+    borderaxespad=0,
+)
+
+f.set_title("Enrichment of most stable core genes", fontsize=16)
```
jjc2718, line 181: These plots might be more informative if they have the same x-axis scale, or at least a more standard scale. In this plot the largest p-value would still be significant at a cutoff of 0.05, but in the other plot the largest p-values (a similar color) are close to 1, which is a bit confusing at a first glance.

3_core_core_analysis/5_KEGG_enrichment_of_stable_genes.py
```diff
@@ -144,17 +146,86 @@ def KEGG_enrichment_of_stable_genes(similarity_score_df, gene_list, kegg_df):
...
+    y="enriched KEGG pathway",
+    s=300,
+    # size="corrected p-value",
+    hue="corrected p-value",
+    # hue_order=pval_lst,
+    # size_order=pval_lst,
+    # sizes=(100, 10),
+    legend="full",
 )
+plt.legend(
```
jjc2718, line 205: Might be good to round these - having this many digits after the decimal doesn't really add much IMO.

3_core_core_analysis/6_explore_secretion_genes.py
```diff
@@ -0,0 +1,126 @@
...
+    alpha=0.8,
+)
+
+plt.title("Stability of secretion system genes", fontsize=14)
+plt.legend()
+# -
+
+# We hypothesized that most secretion machinery genes would be conserved but that secreted proteins (i.e. effector proteins) would be less conserved. In general, the effector proteins were not included in the KEGG annotations, which is probably why these secretion systems were found to be highly stable.
+#
+# There are some low scoring genes found, especially for the T6SS. Based on a manual inspection, these appear to be _vgrG_ genes that do encode some of the effector proteins.
```
jjc2718, line 126: Neat analysis! So this is what you expected to find?

6_common_genes_analysis/1_common_DEGs_and_gene_groups.py
```diff
@@ -419,12 +441,10 @@
...
+# In order to calculate the significance of the observed data, i.e. the total probability of observing data as extreme or more extreme if the null hypothesis is true using the Fisher's exact test.
+# We found that the percentage of PAO1 core genes that are common is ~25% compared to ~34% accessory genes. In PA14, the percentage of core genes that are common is ~25% compared to ~25% accessory genes.
 #
-# * For the Fisher’s exact test, the null hypothesis is that common DEGs are equally likely to be core or accessory genes.
-#     * In order to calculate the significance of the observed data, i.e. the total probability of observing data as extreme or more extreme if the null hypothesis is true
-#     * Percentage of core genes that are common is ~19-20% compared to ~26-32% accessory genes
-#     * There is a negative relationship, meaning if a gene is common DEG, its less likely it’s a core gene
+# Overall, in PAO1 common genes are more likely to be accessory. The genes commonly differentially expressed in response to these stress conditions, are changes that are commonly needed for _P. aeruginosa_ to establish itself within the host.
```
jjc2718, line 448: Do you think this is tied to what you saw in `3_core_core_analysis/5_KEGG_enrichment_of_stable_genes.ipynb`, where the least stable core genes were also (very weakly) enriched for stress response genes? Maybe less stable core genes almost act like accessory genes in some ways?

Just throwing ideas out there, you've probably thought about this much more than I have.

scripts/gene_relationships.py
```diff
@@ -254,6 +309,55 @@ def get_relationship_in_expression_space(
...
+):
+    """
+    This function returns the CI for the
+    get_relationship_in_expression_space()
+    """
+
+    for i in range(num_iter):
+        # Randomly sample `genes_to_consider` with replacement
+        # This will return a sample of the same size as
+        # `genes_to_consider` but with a different random set of genes
```
jjc2718, line 329: "Different" meaning in a different order? Or are you taking a subset of the genes?

scripts/gene_relationships.py
```diff
@@ -200,14 +206,43 @@ def get_relationship_in_expression_space(
...
 
         offset_max = corr_subset.shape[1]
-
         # Note: we are removing genes that are co-operonic first
         # and then finding the most co-expressed genes
-        top_corr_genes = list(corr_subset.loc[gene].nlargest(offset_max).index[1:])
+
+        # If calculating the confidence interval, there will be
+        # duplicate rows due to sampling so we need to select
+        # one of the rows
```
jjc2718, line 214: `.duplicated()` with no arguments will take the first row by default, right? Might be good to mention here, if it could potentially be important.

jjc2718 (approved):
Looks good - just a few comments.

scripts/gene_relationships.py
ajlee21: Yes I'll definitely add this to the comment as this caused a bit of an issue for me when I was coding because I was expecting the function to drop duplicates which is not what its doing.

scripts/gene_relationships.py
ajlee21: Different meaning the I am sampling from the genes_to_consider list and returning a list of genes that is the same size as the genes_to_consider where I sampled with replacement, so its a different subset of genes. I'll update this comment

6_common_genes_analysis/1_common_DEGs_and_gene_groups.py
ajlee21: Yea I that's our guess right now but I still need to think about this some more!

3_core_core_analysis/6_explore_secretion_genes.py
ajlee21: Yea that was nice to see. I still need to dive more into the genes with a low similarity score

3_core_core_analysis/5_KEGG_enrichment_of_stable_genes.py
ajlee21: I agree that makes sense. Thank you for the suggestion!

3_core_core_analysis/1_core_core_relationships_across_strains.py
```diff
@@ -275,66 +256,74 @@ def compare_gene_relationships(gene_mapping_dict, mapping_to, pao1_corr, pa14_co
...
+print(pao1_least_threshold, pao1_most_threshold)
+
 pa14_most_stable = pa14_corr_df.sort_values(
     "Transcriptional similarity across strains", ascending=False
 ).head(round(most_percent * len(pa14_corr_df)))
 pa14_least_stable = pa14_corr_df.sort_values(
     "Transcriptional similarity across strains", ascending=False
 ).tail(round(least_percent * len(pa14_corr_df)))
 
+pa14_most_threshold = pa14_most_stable.iloc[-1][
```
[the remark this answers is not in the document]
ajlee21: Yes, by definition core genes are those genes that are homologous across strains. However based on the venn diagrams there are a couple of core genes that are found in one strain type and not the other just because that core gene wasn't measured in the other strain type or our similarity scores were just slightly different mapping from pao1>pa14 versus pa14>pao1. But there are only 2 cases of this so its not worrisome.

PR #50, opened 2021-11-18
Update core-core and core-acc analyses

ajlee21:
The main changes in this PR include:

1. Add new notebook (`3_core_core_analysis/7_stability_vs_genome_location.ipynb`) to determine if there is a relationship between core gene stability and its consistency in location across strains
2. Update core-acc plot in `5_core_acc_analysis/stable_gene_relationships.ipynb` notebook to be more clear. Just the code to make the plots was updated here.
3. Compare most core genes that are most co-expressed with exoU vs exoS in `5_core_acc_analysis/3_explore_exo_modules.ipynb`
4. Look into what accessory genes are related to least stable genes in `5_core_acc_analysis/least_stable_core_acc_relationships.ipynb`

Additional files changes are from re-running notebooks or re-organizing some of the files to clean up the repo.


3_core_core_analysis/7_stability_vs_genome_location.py
```diff
@@ -0,0 +1,384 @@
...
+#     display_name: Python [conda env:core_acc] *
+#     language: python
+#     name: conda-env-core_acc-py
+# ---
+
+# # Stability vs genome location
+#
+# Here we want to know about the relationship between stability and genome distance: Are more stable genes located in similar locations on the genome in PAO1 vs PA14 compared to least stable genes?
+#
+# We hypothesize that least stable genes are not syntenic (i.e. located in different regions of the genome) across strains, which might indicate a different some transcriptional re-wiring across strain types.
```
danich1, line 21:
```suggestion
# We hypothesize that least stable genes are not syntenic (i.e. located in different regions of the genome) across strains, which might indicate different transcriptional re-wiring across strain types.
```

3_core_core_analysis/7_stability_vs_genome_location.py
```diff
@@ -0,0 +1,384 @@
...
+gene_mapping_pao1 = gene_mapping_pao1.query("num_mapped_genes==1")
+# Select only PA14 column
+gene_mapping_pao1 = gene_mapping_pao1["PA14_ID"].to_frame()
+
+gene_mapping_pao1.head()
+
+
+# ## Approach 1
+#
+# For each most and least stable gene, select the neighboring core genes and determine if the neighboring core genes in PAO1 match with the ones in PA14. If the proprotion of matches is high then that indicates that the most/least stable genes is located in the same relative position on the genome in PAO1 and PA14.
```
danich1, line 86:
```suggestion
# For each most and least stable gene, select the neighboring core genes and determine if the neighboring core genes in PAO1 match with the ones in PA14. If the proprotion of matches is high then that indicates that the most/least stable genes are located in the same relative position on the genome in PAO1 and PA14.
```

My change can go either way: e.g., you are talking about one single gene or you are talking about multiple genes. Assuming the latter, but feel free to correct
```suggestion
# For each most and least stable gene, select the neighboring core genes and determine if the neighboring core genes in PAO1 match with the ones in PA14. If the proprotion of matches is high then that indicates that the most/least stable genes is located in the same relative position on the genome in PAO1 and PA14.
```
```suggestion
# For each most and least stable gene, select the neighboring core genes and determine if the neighboring core genes in PAO1 match with the ones in PA14. If the proprotion of matches is high then that indicates that the most/least stable genes is located in the same relative position on the genome in PAO1 and PA14.
```

5_core_acc_analysis/1_stable_gene_relationships.py
```diff
@@ -425,107 +374,185 @@
...
+        position=positions.position_dodge(0.8),
+        color="darkgrey",
+    )
+    + geom_hline(aes(yintercept=1.0), linetype="dashed")
+    + labs(
+        x="Rank co-expression",
+        y=r"Fold change" + "\n" + "(% acc genes co-express/% acc genes in genome)",
+        title="Who are most/least stable core genes related to (PAO1)",
+    )
+    + theme(
```
danich1, line 439: Could accomplish the same thing with this one liner:
```python
+ theme_seaborn("white")
```

danich1 (approved):
LGTM! Have a few questions and coding tip for the future. Good Work!

danich1:
Also, due to my obliviousness I added all my comments as single comments instead of as a review (all at once). My bad :)

5_core_acc_analysis/1_stable_gene_relationships.py
ajlee21: Nice! Thanks!

3_core_core_analysis/7_stability_vs_genome_location.py
```diff
@@ -0,0 +1,384 @@
...
+    "stability_dist.svg",
+    format="svg",
+    bbox_inches="tight",
+    transparent=True,
+    pad_inches=0,
+    dpi=300,
+)
+# -
+
+# Both approaches yeild similar trends which is good. So the signal is robust. However, both approaches find that there is not an association between stability and if the gene is located in the same location across strain types. We thought that least stable genes might be located in a different location in PAO1 vs PA14 (i.e. we would expect L1 distance to be higher in least stable genes or percent matching homologs to be lower in least stable genes). We instead find that the distances and percent of matching homologs are similar between most and least stable genes.
```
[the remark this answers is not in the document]
ajlee21: Not a dumb question. I'll try to clarify the comments. Basically the goal of this analysis was to determine if least stable genes are located in the same location on the genome in PAO1 and PA14. There were two approaches we used to answer this:

Approach 1: starts with the least stable core gene in PAO1 and selects its nearest core neighbors. Then map that least stable core gene to PA14. Starting with the homologous least stable core gene in PA14 select its nearest core neighbors. Next, since these neighbors are core genes can map between PAO1 and PA14. We compare the two sets of neighbors and determine how many of the neighbors in PAO1 map to the PA14 neighbors. If the percent of neighbors that map are high then this indicates that the starting least stable gene is located a similar location in both PAO1 and PA14 because the neighboring genes are the same.

Approach 2: Takes the genome location defined by the gene ids and scales them to be in the same range since the number of genes in PA14 is more than in PAO1. Then we take the absolute value of the difference between the least stable gene in PAO1 versus PA14. 

So both approaches are trying to answer the same question, we cannot directly calculate distance because the strains have different genome lengths so we have two approximations

PR #51, opened 2021-12-01
Make draft manuscript figures

ajlee21:
This PR adds a notebook to generate manuscript figures (`figure_generation/`). These figures are a first draft. If you can tell me what you understand about figure 1A and how figure 2A and 2B relate. That would be helpful. Also if you have any additional thoughts on how to make these figures more clear given the takeaways below:

Figure 1 takeaway: There are core genes that are stable/consistent across strain types.

Figure 2 takeaway: The least stable core genes are more co-expressed with accessory genes compared to the most stable core genes.


ben-heil (approved):
So sorry that I missed this yesterday!

I looked through the figures before reading your descriptions to see what might be confusing to someone just looking at figures. These comments are more stylistic than content-oriented, so feel free to ignore them as you see fit:
1) I think the colors are overloaded. Blue is used both for low correlation (anticorrelation?) in figure 1 and stability in figure two. If anticorrelation isn't particularly important, you could solve the overloading by just using a single color red-white gradient instead of a two-color one.
2) I think that red means high correlation in both figure 1 and figure 2, but in figure 2 it's unclear what the correlation is. After thinking about it for awhile, I guess it's the correlation between the gene listed on top and the ones on the sides. I guess I didn't process that 'core', 'core', and 'accessory' were all individual genes. Maybe saying 'core gene A', 'core gene B', or something might help?
3) I missed that the first row was what was shown after the arrow in figure 1A. It may be clearer if you take the last row instead of the first one because then the arrow will be pointing to the data used
4) Good color consistency between 1B and 1C! It's immediately obvious that the most and least stable genes from B are the points in 1C
5) I'm not really sure what the bars are in figure 2A. After thinking about it a bit, I realized they show how figure A corresponds to figure B. Adding a y axis label might make that more clear
6) I'm not really sure what the difference between +10 and 10 are in figure B
7) I'm unsure of how many points have a 100% match in figure 1C. Maybe a violin plot would show the distribution better than a jitter plot?

Other comments:
- Your colors swap meanings between 2A and 2B (dark blue is most stable core genes in A and least stable in B)
- Why is  % acc genes co-expressed / % acc genes a fold-change rather than an enrichment?
- There are a different number of co-expression ranks in the two panels of 2B'
- I think the overall logic flow between the two figures is really good. My understanding is "There are genes that have very similar or very dissimilar expression to their homologs in other strains. If we look at the similar ones, they tend to be coexpressed with core genes, and if we look at the dissimilar ones, they tend to be coexpressed with accessory genes"

ajlee21:
Thank you for all these comments, these are really helpful! I'm glad you followed the overall logic. I'll definitely work on editing the figures based on your suggestions, which I think will make the figures much more clear

PR #53, opened 2021-12-16
Redo analysis using latest version of compendia

ajlee21:
There are many files changed in this PR since this PR is re-running all the notebooks using the latest version of the Pseudomonas dataset after our collaborators have finalized QC filtering. We have updated our notebooks to use this latest version to be consistent and reproducible.

No heavy review needed, all these notebooks have been reviewed previously. The results look similar to previously and these can be seen in the figures `figure_generation/output/`. The one notable difference is there is a decrease in signal strength looking at core genes related to accessory genes. After a little digging into the differences between the datasets used it looks like its a difference in the samples included, but a manual spot check into these experiments did not show any obvious reasons.


0_explore_data/samples_with_mixed_acc_expression.ipynb
```diff
@@ -251,17 +233,10 @@
...
-    "* 2 samples ('SRX1491235', 'SRX1491236') testing transcriptomic response of PAO1 under Elevated Temperature: https://www.ncbi.nlm.nih.gov/sra/?term=SRX1491235\n",
-    "* Sample ('SRX1516058') from a study reports the RNA-Seq analysis of bacteriophage PAK_P3 infecting PAK strain of P. aeruginosa: https://www.ncbi.nlm.nih.gov/sra/?term=SRX1516058\n",
-    "* Sample ('SRX1747042') characterzing the in vitro and in vivo transcriptome of PAO1\n",
-    "* Sample ('SRX2662725', 'SRX2662726', 'SRX2662727', 'SRX2662728') characterizing the transcriptional patterns of PAO1 under different culture conditions: https://www.ncbi.nlm.nih.gov/bioproject/PRJNA360055\n",
-    "* 4 samples ('SRX5000019', 'SRX5000020', 'SRX5000025', 'SRX5000026') co-cultured with Aspergillus fumigatus: https://www.ncbi.nlm.nih.gov/sra/?term=SRX5000026\n",
-    "* SRX6981195, SRX389628, grown in a murine chronic wound infection: https://www.ncbi.nlm.nih.gov/sra/?term=SRX69811951\n",
+    "* [SRX6437734](https://www.ncbi.nlm.nih.gov/sra/?term=SRX6437734) and SRX6437785: From an expteriment comparing transcriptional patterns in biofilm vs planktonic growth. This sample is a biofilm sample.\n",
```
jjc2718, line 236:
```suggestion
    "* [SRX6437734](https://www.ncbi.nlm.nih.gov/sra/?term=SRX6437734) and SRX6437785: From an experiment comparing transcriptional patterns in biofilm vs planktonic growth. This sample is a biofilm sample.\n",
```

5_core_acc_analysis/3_explore_exo_modules.py
```diff
@@ -158,14 +158,14 @@
 
 # Core genes highly co-expressed with both exoS and exoU
 exo_core_both = exo_core_corr[
-    (exo_core_corr["corr to exoS"] > 0.4) & (exo_core_corr["corr to exoU"] > 0.4)
+    (exo_core_corr["corr to exoS"] > 0.4) & (exo_core_corr["corr to exoU"] > 0.2)
```
jjc2718: Is there a general reason you're having to tweak some of these cutoffs? Is this because of the decrease in signal strength you mentioned at the top (and by signal strength, do you just mean expression levels are overall lower in core genes)?

jjc2718 (approved):
Looks good!

5_core_acc_analysis/3_explore_exo_modules.py
ajlee21: The signal strength is more referring to the difference in magnitude of the "least stable" bars using the old data (top) vs the new data (bottom). The change in thresholds here was based on a manual inspection and making sure I captured the genes in the two clusters.

image

image

PR #54, opened 2022-01-07
Another analysis reun and update figure 1

ajlee21:
There are two major changes in this PR:

1. Re-ran all scripts using another iteration of the base compendium. The adjustments made can be found in https://github.com/hoganlab-dartmouth/pa-seq-compendia/pull/9. As a result of this re-run there are many files changed, however **no heavy review needed**, all these notebooks have been reviewed previously. The results look similar to previously and these can be seen in the figures figure_generation/output/. 

2. We updated figure 1 to include the thresholding we performed to separate between PAO1-like and PA14-like samples now that Georgia's manuscript is not including this part of the pipeline in her paper. The updated figure 1 along with all the others can be found `figure_generation/output/` for review

3. Added a new validation analysis to compare the most stable core genes found using the RNA-seq vs array compendium. (`3_core_core_analysis/1a_core_core_relationships_in_array_compendium.ipynb`). Looks like there is some consistency but not majority of genes, perhaps because of the difference in how we're processing the data?


1_processing/3_validate_compendia.py
```diff
@@ -227,21 +225,25 @@
...
-    axis_title=pn.element_text(family="sans-serif", size=12),
+    legend_title=pn.element_text(family="sans-serif", size=14),
+    legend_text=pn.element_text(family="sans-serif", size=12),
+    plot_title=pn.element_text(family="sans-serif", size=16),
+    axis_text=pn.element_text(family="sans-serif", size=12),
+    axis_title=pn.element_text(family="sans-serif", size=14),
+)
+fig2 += pn.guides(
+    colour=pn.guide_legend(
+        title="SRA strain type", override_aes={"alpha": 1, "size": 3}
```
taylorreiter, line 244: "SRA strain type" mean the strain as annotated on user-submitted SRA metadata? Doesn't necessarily need to be changed, just wanted to check if my interpretation is correct

taylorreiter (approved):
From `3_core_core_analysis/1a_core_core_relationships_in_array_compendium.ipynb`, I saw you stated, "Looking at the RNA-seq transcriptional similarity of the array only genes (191 genes), most of these genes were just below the threshold for "most stable" (transcriptional similarity = 0.4 - 0.55) according to the distribution plot. However there are some genes that have a lower similarity score (0.2 - 0.3). Why are these stable in microarray but not RNA-seq?"

I have two questions from that statement -- 
1) was your threshold arbitrary?
2) what is the difference between the number of samples and the diversity of samples (broth, groups creating the data/number of studies, etc) between compendia? Could this impact your results?
3) what does linear regression look like between RNA-seq and microarray similarity scores?

Another point of feedback -- for figure S1 you use the same title plot A and B. It might be nice to use Array/RNAseq in the titles for the respective plots they correspond to.

I also left one inline comment, but this LGTM (I skimmed all of your changes, as you stated these nbs have been reviewed previously and many changes looked like plot formatting differences)

1_processing/3_validate_compendia.py
ajlee21: Yes your interpretation is correct here - SRA strain type = strain annotations from SRA metadata

ajlee21:
> From `3_core_core_analysis/1a_core_core_relationships_in_array_compendium.ipynb`, I saw you stated, "Looking at the RNA-seq transcriptional similarity of the array only genes (191 genes), most of these genes were just below the threshold for "most stable" (transcriptional similarity = 0.4 - 0.55) according to the distribution plot. However there are some genes that have a lower similarity score (0.2 - 0.3). Why are these stable in microarray but not RNA-seq?"
> 
> I have two questions from that statement --
> 
> 1. was your threshold arbitrary?

I'm using the same threshold I used for the RNA-seq compendium, which was arbitarily set for the most part

> 2. what is the difference between the number of samples and the diversity of samples (broth, groups creating the data/number of studies, etc) between compendia? Could this impact your results?

Both the RNA-seq and array compendium contain a diverse mix of experiments, though the number of samples in the RNA-seq compendium is using ~500 PA14 samples and ~900 PAO1 samples vs the array compendium is smaller and only contains ~100 PA14 samples and ~

> 3. what does linear regression look like between RNA-seq and microarray similarity scores?

Very good point. I think it'll be helpful to see what the correlation looks like between the similarity scores for array vs RNA-seq

> Another point of feedback -- for figure S1 you use the same title plot A and B. It might be nice to use Array/RNAseq in the titles for the respective plots they correspond to.

Good point, I might remove one of the distribution plots since its in one of the figures already.

> I also left one inline comment, but this LGTM (I skimmed all of your changes, as you stated these nbs have been reviewed previously and many changes looked like plot formatting differences)

PR #55, opened 2022-02-17
Update readme

ajlee21:
This PR updates the readme file in preparation for manuscript submission. Specifically, the text description of the project was updated to reflect the analysis we performed. Previously our description focused more on the clinical application, which our analysis doesn't actually go into.


taylorreiter (approved):
this looks great!

PR #59, opened 2022-06-15
Additional exploration

ajlee21:
Did some additional exploration:

For the first analysis the question is: are gene level differences in aromatic amino acid genes in PAO1 vs PA14 consistent across the RNA-seq and array compendium? The approach we took was to compare the median expression of aromatic amino acid genes different in PAO1 vs PA14 using RNA-seq and array compendia. We found that the expression activity of AA genes have a similar distribution in PAO1 and PA14. This similar distribution is observed in both RNA-seq and array data

For the second analysis, PCA based on least stable core genes among clinical isolates, PAO1 and PA14. Do clinical isolates cluster with PAO1 or PA14? We found that a few clinical isolates appear to form their own separate cluster distinct from PAO1 and PA14. However, most clinical isolates are in the same cluster as PAO1 and PA14, indicating perhaps that the least stable core genes have similar profiles across strain types despite their neighbors being different across strains (PA14 reference looks the same)


ajlee21:
This branch includes some additional exploration analysis that doesn't fit into the associated manuscript, so I'm closing it

PR #61, opened 2022-06-24
Update figure 1

ajlee21:
This PR makes minor updates to figure 1 based on group discussion:

1. Log-scale clustering of gene expression data
2. Add PA14 PCA plot as panel


0_explore_data/cluster_by_accessory_gene.py
```diff
@@ -568,12 +568,12 @@
 )
 fig4 += pn.scales.scale_x_log10()
 fig4 += pn.scales.scale_y_log10()
-fig4 += pn.geom_point(pn.aes(color="Strain type_pao1"), alpha=0.4)
+fig4 += pn.geom_point(pn.aes(color="Strain type_pao1"), alpha=0.3, size=3, stroke=0.8)
```
taylorreiter, line 571: would it make sense to do an alpha of ~0.5? The points are super faint. Usually I would say dropping it is good because the points along the axes are so dense, but because these aren't the focus, and there isn't tons of overlap for the ones that don't overlap very much, maybe slightly higher color saturation could be nice? Feel free to ignore this if you want.

1_processing/3_validate_compendia.py
```diff
@@ -185,12 +185,14 @@
...
 )
+fig1 += pn.scales.scale_x_log10()
+fig1 += pn.scales.scale_y_log10()
 fig1 += pn.geom_point(pn.aes(color="Strain type"), alpha=0.3, size=3, stroke=0.8)
 fig1 += pn.scale_color_manual(values=colors)
 fig1 += pn.labs(
     x="median expression of PAO1-only genes",
     y="median expression of PA14-only genes",
-    title="PAO1 compendium accessory gene expression",
+    title="PAO1 sample compendium accessory gene expression",
```
taylorreiter, line 195: small change but i think little things like this will make it a lot more clear what set of samples is being analyzed.

taylorreiter (approved):
LGTM!

PR #62, opened 2022-08-01
Manuscript revisions

ajlee21:
This PR contains the following minor revisions in preparation for manuscript submission:

1. Update figure 1 point saturation to be darker so that pao1 vs pa14 points are more distinguishable. With the lighter saturation, readers could think that several pao1's stacked on top of each other could look like a pa14 point.
2. Printed counts to add to the text of the manuscript to clarify and make things more explicit.


taylorreiter (approved):
Oooh I love the printed counts, that's a really nice tie in with the manuscript text. Nice change on the color saturation as well, lgtm!

PR #63, opened 2022-08-15
Combine Figures S2 and S3 

ajlee21:
Small PR to combine Supplemental Figures S2 and S3 since we are limited to 10 supplemental materials


taylorreiter (approved):
LGTM!

